PR #1, opened 2020-03-17
Review ABI vNEXT.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

### DIFF
--- a/abi-versions/README.md
+++ b/abi-versions/README.md
@@ -1,0 +1,40 @@
+# Proxy-Wasm ABI specification
+
+## ABI versioning
+
+The ABI is in the form of: `<major>.<minor>.<patch>`
+
+Breaking changes require increasing the major version.
+
+Adding new functions requires increasing the minor version.
+
+Adding new enum values requires increasing the patch version.
+
+Wasm modules should export `proxy_abi_version_<major>_<minor>_<patch>` function, indicating which
+version of the ABI they were compiled against.
+
+Host environments should be able to support multiple versions of the ABI.
+
+## Naming conventions
+
+All functions are using `snake_case`, and they start with the `proxy_` prefix. Implementation
+specific extensions should use different prefixes (e.g. `envoy_`, `nginx_`, `ats_`, etc.) in order
+to avoid polluting the common namespace. However, this is highly discouraged, and should be used
+only for implementing optional proxy-specific functionality, otherwise we’re going to lose
+portability.
+
+Function names are in the form of: `proxy_<verb>_<namespace>_<field>_<field>` (e.g.
+`proxy_on_http_request_headers`, `proxy_get_map_value`, etc).
+
+## Memory ownership
+
+No memory-managed data is passed to Wasm module in any of the event callbacks, and Wasm module must
+request memory-managed data from the host environment if it’s interested in it, e.g.
+`proxy_on_http_request_body` only passes the information about the number of available request body
+bytes, but the module must request them using `proxy_get_buffer`. When that happens, the host
+environment will request the Wasm module to allocate memory using `proxy_on_memory_allocate` and
+copy requested data into it. In doing so, it passes memory ownership of that data to the Wasm
+module, which is expected to free this memory when it no longer needs it.
+
+SDKs implementing this ABI are expected to automatically handle freeing of the memory-managed data
+for the developers.
--- a/abi-versions/vNEXT/README.md
+++ b/abi-versions/vNEXT/README.md
@@ -327,8 +327,7 @@ Note: The same unique context identifier (`stream_id`) is shared by a pair of
 Called when HTTP request headers for stream `stream_id` are received from
 downstream.
 
-Headers (`num_headers` entries) can be retrieved using `proxy_get_map` and/or
-`proxy_get_map_value`.
+Headers (`num_headers` entries) can be retrieved using `proxy_get_map_values`.
 
 Return value (`next_action`) instructs the host environment what action to take:
 - `Continue` means that this HTTP request should be accepted and that HTTP
@@ -385,8 +384,7 @@ Return value (`next_action`) instructs the host environment what action to take:
 Called when HTTP request trailers for stream `stream_id` are received from
 downstream.
 
-Trailers (`num_trailers` entries) can be retrieved using `proxy_get_map` and/or
-`proxy_get_map_value`.
+Trailers (`num_trailers` entries) can be retrieved using `proxy_get_map_values`.
 
 Return value (`next_action`) instructs the host environment what action to take:
 - `Continue` means that HTTP trailers should be forwarded upstream.
@@ -409,8 +407,7 @@ Return value (`next_action`) instructs the host environment what action to take:
 Called for each HTTP/2 METADATA frame for stream `stream_id` received from
 downstream.
 
-Metadata (`num_elements` entries) can be retrieved using `proxy_get_map` and/or
-`proxy_get_map_value`.
+Metadata (`num_elements` entries) can be retrieved using `proxy_get_map_values`.
 
 Return value (`next_action`) instructs the host environment what action to take:
 - `Continue` means that HTTP/2 METADATA elements should be forwarded upstream.
@@ -434,8 +431,7 @@ Return value (`next_action`) instructs the host environment what action to take:
 Called when HTTP response headers for stream `stream_id` are received from
 upstream.
 
-Headers (`num_headers` entries) can be retrieved using `proxy_get_map` and/or
-`proxy_get_map_value`.
+Headers (`num_headers` entries) can be retrieved using `proxy_get_map_values`.
 
 Return value (`next_action`) instructs the host environment what action to take:
 - `Continue` means that HTTP headers should be forwarded downstream.
@@ -489,8 +485,7 @@ Return value (`next_action`) instructs the host environment what action to take:
 Called when HTTP response trailers for stream `stream_id` are received from
 upstream.
 
-Trailers (`num_trailers` entries) can be retrieved using `proxy_get_map` and/or
-`proxy_get_map_value`.
+Trailers (`num_trailers` entries) can be retrieved using `proxy_get_map_values`.
 
 Return value (`next_action`) instructs the host environment what action to take:
 - `Continue` means that HTTP trailers should be forwarded downstream.
@@ -513,8 +508,7 @@ Return value (`next_action`) instructs the host environment what action to take:
 Called for each HTTP/2 METADATA frame for stream `stream_id` received from
 upstream.
 
-Metadata (`num_elements` entries) can be retrieved using `proxy_get_map` and/or
-`proxy_get_map_value`.
+Metadata (`num_elements` entries) can be retrieved using `proxy_get_map_values`.
 
 Return value (`next_action`) instructs the host environment what action to take:
 - `Continue` means that HTTP/2 METADATA elements should be forwarded downstream.
@@ -534,9 +528,9 @@ Return value (`next_action`) instructs the host environment what action to take:
   - `i32 (uint32_t) response_code`
   - `i32 (const char*) response_code_details_data`
   - `i32 (size_t) response_code_details_size`
-  - `i32 (const char*) response_body_data`
+  - `i32 (iovec_array_t) response_body_data`
   - `i32 (size_t) response_body_size`
-  - `i32 (const char*) additional_headers_map_data`
+  - `i32 (iovec_array_t[]) additional_headers_map_data`
   - `i32 (size_t) additional_headers_size`
   - `i32 (uint32_t) grpc_status`
 * returns:
@@ -595,14 +589,15 @@ Returns `InvalidOperation` when stream is already closed.
   - `i32 (proxy_buffer_type_t) buffer_type`
   - `i32 (size_t) offset`
   - `i32 (size_t) max_size`
-  - `i32 (const char**) return_buffer_data`
+  - `i32 (iovec_array_t*) return_buffer_data`
   - `i32 (size_t*) return_buffer_size`
 * returns:
   - `i32 (proxy_result_t) call_result`
 
 Returns content of the buffer `buffer_type`. At most `max_size` bytes will be
-returned, starting from the `offset`. Returned bytes are written into provided
-memory slice (`return_buffer_data`, `return_buffer_size`).
+returned, starting from the `offset`. Returned bytes are written into memory
+slices pointed by a list of vectors (`return_buffer_data`,
+`return_buffer_size`).
 
 Returns `Ok` on success.
 
@@ -619,15 +614,16 @@ Returns `InvalidMemoryAccess` when `return_buffer_data` and/or
   - `i32 (proxy_buffer_type_t) buffer_type`
   - `i32 (offset_t) offset`
   - `i32 (size_t) size`
-  - `i32 (const char*) buffer_data`
+  - `i32 (iovec_array_t) buffer_data`
   - `i32 (size_t) buffer_size`
 * returns:
   - `i32 (proxy_result_t) call_result`
 
-Sets content of the buffer `buffer_type` to the provided value (`buffer_data`,
-`buffer_size`), replacing `size` bytes in the existing buffer, starting at the
-`offset`. Host environment must adjust the size of the destination buffer
-(e.g. shrink the buffer if target `size` is larger than `buffer_size`).
+Sets content of the buffer `buffer_type` to the values from memory slices
+pointed by the provided list of vectors (`buffer_data`, `buffer_size`),
+replacing `size` bytes in the existing buffer, starting at the `offset`. Host
+environment must adjust the size of the destination buffer (e.g. shrink the
+buffer if target `size` is larger than `buffer_size`).
 
 Returns `Ok` on success.
 
@@ -636,83 +632,57 @@ Returns `BadArgument` for unknown `buffer_type`.
 Returns `InvalidMemoryAccess` when `buffer_data` and/or `buffer_size` point to
 invalid memory address.
 
-#### `proxy_get_map`
-
-* params:
-  - `i32 (proxy_map_type_t) map_type`
-  - `i32 (const char**) return_map_data`
-  - `i32 (size_t*) return_map_size`
-* returns:
-  - `i32 (proxy_result_t) call_result`
-
-Returns all key-value pairs from the given map (`map_type`).
-
-Returns `Ok` on success.
-
-Returns `BadArgument` for unknown `map_type`.
-
-Returns `Empty` when `map_type` doesn't contain any data.
-
-Returns `InvalidMemoryAccess` when `return_map_data` and/or `return_map_size`
-point to invalid memory address.
-
-#### `proxy_set_map`
-
-* params:
-  - `i32 (proxy_map_type_t) map_type`
-  - `i32 (const char*) map_data`
-  - `i32 (size_t) map_size`
-* returns:
-  - `i32 (proxy_result_t) call_result`
-
-Sets all key-value pairs in the given map (`map_type`).
-
-Returns `Ok` on success.
-
-Returns `BadArgument` for unknown `map_type`.
-
-Returns `InvalidMemoryAccess` when `map_data` and/or `map_size` point to invalid
-memory address.
-
 #### `proxy_get_map_values`
 
 * params:
   - `i32 (proxy_map_type_t) map_type`
-  - `i32 (const char*) key_data`
-  - `i32 (size_t) key_size`
-  - `i32 (const char**) return_values_data`
-  - `i32 (size_t*) return_values_size`
+  - `i32 (iovec_array_t) keys_data`
+  - `i32 (size_t) keys_size`
+  - `i32 (iovec_array_t[]*) return_map_data`
+  - `i32 (size_t*) return_map_size`
 * returns:
   - `i32 (proxy_result_t) call_result`
 
-Returns content of the key (`key_data`, `key_size`) from the map (`map_type`).
+Returns map containing key-values for keys from the provided list of vectors
+(`keys_data`, `keys_size`) from the map (`map_type`) in a list of lists
+of vectors (`return_map_data`, `return_map_size`) in which the first element
+in each list represents a key, followed by a zero or more values.
+
+If the list of keys is empty, then all key-values in the map are returned.
+
+In context of HTTP header maps, the special key `"*"` can be used to retrieve
+all regular HTTP headers, and the special key `":*"` can be used to retrieve
+all pseudo-headers.
 
 Returns `Ok` on success.
 
 Returns `BadArgument` for unknown `map_type`.
-
-Returns `NotFound` when key (`key_data`, `key_size`) doesn't exist.
 
 Returns `InvalidMemoryAccess` when `key_data`, `key_size`, `return_values_data`
 and/or `return_values_size` point to invalid memory address.
 
-#### `proxy_set_map_key_values`
+#### `proxy_set_map_values`
 
 * params:
   - `i32 (proxy_map_type_t) map_type`
-  - `i32 (const char*) key_data`
-  - `i32 (size_t) key_size`
-  - `i32 (const char*) values_data`
-  - `i32 (size_t) values_size`
-  - `i32 (bool) multiple_values`
+  - `i32 (iovec_array_t) remove_keys_data`
+  - `i32 (size_t) remove_keys_size`
+  - `i32 (iovec_array_t[]) map_data`
+  - `i32 (size_t) map_size`
 * returns:
   - `i32 (proxy_result_t) call_result`
 
-Sets content of the key (`key_data`, `key_size`) in the map (`map_type`) to the
-provided values (`values_data`, `values_size`).
+Adds key-values from the provided list of lists of vectors (`map_data`,
+`map_size`) to the given map (`map_type`). The first element in each list
+represents a key, followed by a zero or more values. For the existing keys,
+new values will be added to the existing ones.
 
-If `multiple_values` is `false`, then `values_data`, `values_size` contains
-a single value, otherwise it contains multiple values serialized as a list.
+Keys to remove and/or which values will be replaced can be provided in an
+optional list of vectors (`remove_keys_data`, `remove_keys_size`).
+
+In context of HTTP header maps, the special key `"*"` can be used to remove
+all regular HTTP headers, and the special key `":*"` can be used to remove
+all pseudo-headers.
 
 Returns `Ok` on success.
 
@@ -720,49 +690,6 @@ Returns `BadArgument` for unknown `map_type`.
 
 Returns `InvalidMemoryAccess` when `key_data`, `key_size`, `values_data` and/or
 `values_size` point to invalid memory address.
-
-#### `proxy_add_map_key_values`
-
-* params:
-  - `i32 (proxy_map_type_t) map_type`
-  - `i32 (const char*) key_data`
-  - `i32 (size_t) key_size`
-  - `i32 (const char*) values_data`
-  - `i32 (size_t) values_size`
-  - `i32 (bool) multiple_values`
-* returns:
-  - `i32 (proxy_result_t) call_result`
-
-Adds new values (`values_data`, `values_size`) to the key (`key_data`,
-`key_size`) in the map (`map_type`).
-
-If `multiple_values` is `false`, then `values_data`, `values_size` contains
-a single value, otherwise it contains multiple values serialized as a list.
-
-Returns `Ok` on success.
-
-Returns `BadArgument` for unknown `map_type`.
-
-Returns `InvalidMemoryAccess` when `key_data`, `key_size`, `values_data` and/or
-`values_size` point to invalid memory address.
-
-#### `proxy_remove_map_key`
-
-* params:
-  - `i32 (proxy_map_type_t) map_type`
-  - `i32 (const char*) key_data`
-  - `i32 (size_t) key_size`
-* returns:
-  - `i32 (proxy_result_t) call_result`
-
-Removes the key (`key_data`, `key_size`) from the map (`map_type`).
-
-Returns `Ok` on success.
-
-Returns `BadArgument` for unknown `map_type`.
-
-Returns `InvalidMemoryAccess` when `key_data` and/or `key_size` point to invalid
-memory address.
 
 
 ## Shared key-value store
@@ -804,7 +731,7 @@ and/or `return_kvstore_id` point to invalid memory address.
   - `i32 (uint32_t) kvstore_id`
   - `i32 (const char*) key_data`
   - `i32 (size_t) key_size`
-  - `i32 (const char**) return_values_data`
+  - `i32 (iovec_array_t*) return_values_data`
   - `i32 (size_t*) return_values_size`
   - `i32 (uint32_t*) return_cas`
 * returns:
@@ -831,18 +758,14 @@ Returns `InvalidMemoryAccess` when `key_data`, `key_size`, `return_values_data`,
   - `i32 (uint32_t) kvstore_id`
   - `i32 (const char*) key_data`
   - `i32 (size_t) key_size`
-  - `i32 (const char*) values_data`
+  - `i32 (iovec_array_t) values_data`
   - `i32 (size_t) values_size`
-  - `i32 (bool) multiple_values`
   - `i32 (uint32_t) cas`
 * returns:
   - `i32 (proxy_result_t) call_result`
 
 Sets the values of the key (`key_data`, `key_size`) in a shared key-value store
 (`kvstore_id`) to the values (`values_data`, `values_size`).
-
-If `multiple_values` is `false`, then `values_data`, `values_size` contains
-a single value, otherwise it contains multiple values serialized as a list.
 
 If compare-and-switch value (`cas`) is provided and non-zero, then it must match
 the current value in order for update to succeed.
@@ -863,18 +786,14 @@ Returns `InvalidMemoryAccess` when `key_data`, `key_size`, `values_data` and/or
   - `i32 (uint32_t) kvstore_id`
   - `i32 (const char*) key_data`
   - `i32 (size_t) key_size`
-  - `i32 (const char*) values_data`
+  - `i32 (iovec_array_t) values_data`
   - `i32 (size_t) values_size`
-  - `i32 (bool) multiple_values`
   - `i32 (uint32_t) cas`
 * returns:
   - `i32 (proxy_result_t) call_result`
 
 Adds new values (`values_data`, `values_size`) to the key (`key_data`,
 `key_size`) in a shared key-value store (`kvstore_id`).
-
-If `multiple_values` is `false`, then `values_data`, `values_size` contains
-a single value, otherwise it contains multiple values serialized as a list.
 
 If compare-and-switch value (`cas`) is provided and non-zero, then it must match
 the current value in order for update to succeed.
@@ -965,7 +884,7 @@ Returns `InvalidMemoryAccess` when `queue_name_data`, `queue_name_size` and/or
 
 * params:
   - `i32 (uint32_t) queue_id`
-  - `i32 (const char**) return_payload_data`
+  - `i32 (iovec_array_t*) return_payload_data`
   - `i32 (size_t*) return_payload_size`
 * returns:
   - `i32 (proxy_result_t) call_result`
@@ -984,7 +903,7 @@ Returns `InvalidMemoryAccess` when `return_payload_data` and/or
 
 * params:
   - `i32 (uint32_t) queue_id`
-  - `i32 (const char*) payload_data`
+  - `i32 (iovec_array_t) payload_data`
   - `i32 (size_t) payload_size`
 * returns:
   - `i32 (proxy_result_t) call_result`
@@ -1172,11 +1091,11 @@ Returns `BadArgument` when the metric (`metric_id`) doesn't exist.
 * params:
   - `i32 (const char*) upstream_name_data`
   - `i32 (size_t) upstream_name_size`
-  - `i32 (const char*) headers_map_data`
+  - `i32 (iovec_array_t[]) headers_map_data`
   - `i32 (size_t) headers_map_size`
-  - `i32 (const char*) body_data`
+  - `i32 (iovec_array_t) body_data`
   - `i32 (size_t) body_size`
-  - `i32 (const char*) trailers_map_data`
+  - `i32 (iovec_array_t[]) trailers_map_data`
   - `i32 (size_t) trailers_map_size`
   - `i32 (uint32_t) timeout_milliseconds`
   - `i32 (uint32_t*) return_callout_id`
@@ -1214,7 +1133,7 @@ invalid memory address.
 Called when the complete response to the HTTP call (`callout_id`) is received.
 
 Headers (`num_headers` entries) and trailers (`num_trailers` entries) can be
-retrieved using `proxy_get_map` and/or `proxy_get_map_value`.
+retrieved using `proxy_get_map_values`.
 
 Response body (of size `body_size`) can be retrieved using `proxy_get_buffer`.
 
@@ -1232,9 +1151,9 @@ Response body (of size `body_size`) can be retrieved using `proxy_get_buffer`.
   - `i32 (size_t) service_name_size`
   - `i32 (const char*) method_data`
   - `i32 (size_t) method_size`
-  - `i32 (const char*) initial_metadata_map_data`
+  - `i32 (iovec_array_t[]) initial_metadata_map_data`
   - `i32 (size_t) initial_metadata_map_size`
-  - `i32 (const char*) grpc_message_data`
+  - `i32 (iovec_array_t) grpc_message_data`
   - `i32 (size_t) grpc_message_size`
   - `i32 (uint32_t) timeout_milliseconds`
   - `i32 (uint32_t*) return_callout_id`
@@ -1270,7 +1189,7 @@ Returns `InvalidMemoryAccess` when `grpc_service_data`, `grpc_service_size`,
   - `i32 (size_t) service_name_size`
   - `i32 (const char*) method_data`
   - `i32 (size_t) method_size`
-  - `i32 (const char*) initial_metadata_map_data`
+  - `i32 (iovec_array_t[]) initial_metadata_map_data`
   - `i32 (size_t) initial_metadata_map_size`
   - `i32 (uint32_t*) return_callout_id`
 * returns:
@@ -1294,7 +1213,7 @@ Returns `InvalidMemoryAccess` when `grpc_service_data`, `grpc_service_size`,
 
 * params:
   - `i32 (uint32_t) callout_id`
-  - `i32 (const char*) grpc_message_data`
+  - `i32 (iovec_array_t) grpc_message_data`
   - `i32 (size_t) grpc_message_size`
 * returns:
   - `i32 (proxy_result_t) call_result`
@@ -1341,15 +1260,15 @@ Returns `BadArgument` when the gRPC call (`callout_id`) doesn't exist.
 
 * params:
   - `i32 (uint32_t) callout_id`
-  - `i32 (size_t) num_elements`
+  - `i32 (size_t) num_headers`
 * returns:
   - none
 
 Called when header metadata in the response to the gRPC call (`callout_id`) is
 received.
 
-Header metadata (`num_headers` entries) can be retrieved using `proxy_get_map`
-and/or `proxy_get_map_value`.
+Header metadata (`num_headers` entries) can be retrieved using
+`proxy_get_map_values`.
 
 #### `proxy_on_grpc_call_response_message`
 
@@ -1369,15 +1288,15 @@ Content of the gRPC message (of size `data_size`) can be retrieved using
 
 * params:
   - `i32 (uint32_t) callout_id`
-  - `i32 (size_t) num_elements`
+  - `i32 (size_t) num_trailers`
 * returns:
   - none
 
 Called when trailer metadata in the response to the gRPC call (`callout_id`) is
 received.
 
-Trailer metadata (`num_headers` entries) can be retrieved using `proxy_get_map`
-and/or `proxy_get_map_value`.
+Trailer metadata (`num_trailers` entries) can be retrieved using
+`proxy_get_map_values`.
 
 #### `proxy_on_grpc_call_close`
 
@@ -1417,9 +1336,9 @@ custom callback. Returning `0` indicates unknown `custom_callback_id`.
 
 * params:
   - `i32 (uint32_t) custom_function_id`
-  - `i32 (const char *) parameters_data`
+  - `i32 (iovec_array_t) parameters_data`
   - `i32 (size_t) parameters_size`
-  - `i32 (const char**) return_results_data`
+  - `i32 (iovec_array_t*) return_results_data`
   - `i32 (size_t*) return_results_size`
 * returns:
   - `i32 (proxy_result_t) call_result`
@@ -1440,23 +1359,35 @@ address.
 
 The encoding of integers is little endian.
 
-### List of values.
+### Vector (`iovec_t`)
 
-List of values is serialized as: 32-bit number of elements in the list,
-followed by a series of 32-bit lengths of each value, followed by a series of
-payload of each value.
+A vector is represented by a pair of 32-bit integers. First number points to the
+address in memory where the value's data starts at, second number represents its
+length.
 
-e.g. `{"a", "b"}` is serialized as: `{ 0x2, 0x1, 0x1, "a", "b" }`.
+e.g. vector pointing to `"one"` would be represented by `{0xAA, 0x03}`, assuming
+that `"one"` starts at address `0xAA`.
 
-### Map with multiple values.
+### List of vectors (`iovec_array_t`)
 
-Map with multiple values is serialized as: 32-bit number of keys in the map,
-followed by a series of pairs of 32-bit lengths of each key and its serialized
-list of values, followed by a series of key names and its serialized list of
-values.
+A list of vectors is serialized as a continuous list of `iovec_t` elements.
 
-e.g. `{"key": {"a", "b"}}` is serialized as:
-`{ 0x1, 0x3, 0x5, "k", "e", "y", 0x2, 0x1, 0x1, "a", "b" }`.
+e.g. a list of 2 vectors pointing to `{"one", "two"}` would be serialized as
+`{{0xAA, 0x03}, {0xBB, 0x03}}`, assuming that `"one"` starts at address `0xAA`,
+and `"two"` starts at address `0xBB`.
+
+### List of lists of vectors (`iovec_array_t[]`)
+
+A list of lists of vectors is serialized as a continuous list of
+`iovec_array_t`, where each top-level vector points to a list of vectors.
+
+e.g. a list of lists of vectors pointing to `{{"key", "a", "b"}, {"empty"}}`
+would be serialized as `{{0xEE, 0x03}, {0xFF, 0x01}}`, where `0xEE` contains
+`{{0xCC, 0x03}, {0xAA, 0x01}, {0xBB, 0x01}}`, which represents a list of 3
+vectors: `"key"`, `"a"`, `"b"`, assuming that `"key"` starts at address `0xCC`,
+`"a"` starts at address `0xAA`, and `"b"` starts at address `0xBB`, and `0xFF`
+contains `{{0xDD, 0x05}}`, which represents a list of 1 vector: `"empty"`,
+assuming that `"empty"` starts at address `0xDD`.
 
 
 # Built-in types
--- a/abi-versions/vNEXT/README.md
+++ b/abi-versions/vNEXT/README.md
@@ -1,0 +1,919 @@
+# Proxy-Wasm vNEXT ABI specification
+
+# Functions implemented in the Wasm module
+
+All functions implemented in the Wasm module, other than the integration and memory management
+functions, include context identifier (`context_id`) as the first parameter, which should be used to
+distinguish between different contexts.
+
+
+## Integration
+
+### `_start`
+
+* params:
+  - none
+* returns:
+  - none
+
+Start function which is called when the module is loaded and initialized. This can be used by SDKs
+to setup and/or initialize state, but no proxy_ functions can be used at that point yet.
+
+
+### `proxy_abi_version_X_Y_Z`
+
+* params:
+  - none
+* returns:
+  - none
+
+Exports ABI version (vX.Y.Z) this module was compiled against.
+
+Considered alternatives: Ideally, this would be exported as a global variable, but existing
+toolchains have trouble with that (some don’t support global variables at all, some incorrectly
+export language-specific and not Wasm-generic variables).
+
+
+## Memory management
+
+### `proxy_on_memory_allocate`
+
+* params:
+  - `i32 (size_t) memory_size`
+* returns:
+  - `i32 (void*) allocated_ptr`
+
+Allocates memory using in-VM memory allocator and returns it to the host.
+
+
+## Module lifecycle
+
+### `proxy_on_context_create`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (uint32_t) parent_context_id`
+* returns:
+  - none
+
+Called when the host environment creates a new root context (if `parent_context_id` is `0`) or a new
+per-stream context.
+
+
+### `proxy_on_done`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - `i32 (bool) is_done`
+
+Called when the host environment is done processing the context (`context_id`). Return value
+indicates when the Wasm VM is done with the processing as well.
+
+
+### `proxy_on_log`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - none
+
+Called in the logging phase.
+
+
+### `proxy_on_delete`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - none
+
+Called when the host environment removes the context (`context_id`). This is used to signal that VM
+should stop tracking that `context_id` and remove all associated state.
+
+
+## Configuration
+
+### `proxy_on_vm_start`
+
+* params:
+  - `i32 (uint32_t) root_context_id`
+  - `i32 (size_t) vm_configuration_size`
+* returns:
+  - `i32 (bool) success`
+
+Called when the host environment starts the WebAssembly Virtual Machine. Its configuration
+(`vm_configuration_size`) might be retrieved using `proxy_get_buffer`.
+
+
+### `proxy_on_configure`
+
+* params:
+  - `i32 (uint32_t) root_context_id`
+  - `i32 (size_t) plugin_configuration_size`
+* returns:
+  - `i32 (bool) success`
+
+Called when the host environment starts the plugin. Its configuration (`plugin_configuration_size`)
+might be retrieved using `proxy_get_buffer`.
+
+
+## Timers
+
+### `proxy_on_tick`
+
+* params:
+  - `i32 (uint32_t) root_context_id`
+* returns:
+  - none
+
+Timer called every tick period. Tick period can be configured using
+`proxy_set_tick_period_milliseconds`.
+
+
+## TCP/UDP/QUIC stream (L4) extensions
+
+Note: `downstream` means the connection between client and proxy, and `upstream` means the
+connection between proxy and backend (aka “next hop”).
+
+### `proxy_on_new_connection`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - `i32 (proxy_action_t) next_action`
+
+Called on a new TCP connection.
+
+
+### `proxy_on_downstream_data`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (size_t) data_size`
+  - `i32 (bool) end_of_stream`
+* returns:
+  - `i32 (proxy_action_t) next_action`
+
+Called for each data chunk received from downstream.
+
+
+### `proxy_on_downstream_close`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (proxy_peer_type_t) peer_type`
+* returns:
+  - none
+
+Called when downstream connection is closed.
+
+
+### `proxy_on_upstream_data`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (size_t) data_size`
+  - `i32 (bool) end_of_stream`
+* returns:
+  - `i32 (proxy_action_t) next_action`
+
+Called for each data chunk received from upstream.
+
+
+### `proxy_on_upstream_close`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (proxy_peer_type_t) peer_type`
+* returns:
+  - none
+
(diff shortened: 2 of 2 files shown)


abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+  - `i32 (uint32_t*) return_queue_id`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Register a shared queue using a given name (`queue_name_data`, `queue_name_size`). It can be
+referred to in `proxy_enqueue_shared_queue` and `proxy_dequeue_shared_queue` using returned unique
+queue identifier (`queue_id`).
+
+
+### `proxy_resolve_shared_queue`
```
mattklein123: I commented on this in the C++ PR, but why do we need both this function as well as `proxy_register_shared_queue`? Could we have a single function with a bool param `create_if_not_exists` or something like that? Otherwise the functionality is the same?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+* params:
+  - `i32 (uint32_t) queue_id`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Remove queue (`queue_id`).
+
+
+## Stats/metrics
```
mattklein123: I've commented on this elsewhere but I *very* strongly recommend that you split counters/gauges/histograms into independent APIs. The reason is that they are all subtly different. If we don't want to do this, I think there are a few things we need to do here:
1) Better clarify what call patterns will result in invalid arguments (incrementing a counter with a negative offset, trying to set a counter value, etc.).
2) Figure out how to deal with histograms correctly because `proxy_get_metric` is not going to work well for them.

mattklein123:
cc @htuch to take a look at this also.

abi-versions/README.md
```diff
@@ -0,0 +1,40 @@
...
+`proxy_on_http_request_headers`, `proxy_get_map_value`, etc).
+
+## Memory ownership
+
+No memory-managed data is passed to Wasm module in any of the event callbacks, and Wasm module must
+request memory-managed data from the host environment if it’s interested in it, e.g.
+`proxy_on_http_request_body` only passes the information about the number of available request body
+bytes, but the module must request them using `proxy_get_buffer`. When that happens, the host
+environment will request the Wasm module to allocate memory using `proxy_on_memory_allocate` and
+copy requested data into it. In doing so, it passes memory ownership of that data to the Wasm
```
jmarantz: This seems unfortunate if I want to write a high-performance WASM filter that examines or modifies content.

What do you you think of supporting, in the context of proxy_on_http_request_body, some APIs to enable a filter to access the body in various ways without copying all of it? E.g.

 * copying ranges of bytes into the WASM address space
 * searching for bytes/substrings
 * global-replace
 * global-replace-RE2
 * parse as JSON and provide JSON API to content
 * send content to a socket without brining it into the WASM space

Maybe not all of these should be done at first (or ever) but establishing this pattern seems useful if you want to write performance content-aware WASM filters.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+## Module lifecycle
+
+### `proxy_on_context_create`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (uint32_t) parent_context_id`
+* returns:
+  - none
```
yuval-k: I had  a [discussion in the architecture doc](https://docs.google.com/document/d/1YPbxuEIPwKsW4Ft73diHRBTUNbYIlrQHNnS9dXdn778/edit?disco=AAAAEHXWueQ) with @jplevyak on this function, that I would  like to get the opinion of the reviewers on.

i think it would be useful to have a way to allow runtimes use a void* coming from the runtime as their context id, for the following reasons:
- this id is used to retrieve the context on every request. this means that the runtime essentially has to store these in a map<id, context> for fast access. some runtimes (tiny-go comes to  mind) may not have an efficient map implementation.
- it will make runtimes simpler\easier to write

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+  - `i32 (uint32_t) root_context_id`
+  - `i32 (size_t) vm_configuration_size`
+* returns:
+  - `i32 (bool) success`
+
+Called when the host environment starts the WebAssembly Virtual Machine. Its configuration
+(`vm_configuration_size`) might be retrieved using `proxy_get_buffer`.
+
+
+### `proxy_on_configure`
```
yuval-k: the previous api had `proxy_validate_configuration` for the use case of validating configuration from the control plane. is the plan to use`proxy_on_configure` for both  cases?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+### `proxy_set_effective_context`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Change the effective context to `context_id`. This function is usually used to change the context
+after receiving `proxy_on_http_call_response`, `proxy_on_grpc_call_response` or
+`proxy_on_queue_ready`.
```
yuval-k: is this needed because these callbacks are called in the context of the root context?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+  - `i32 (offset_t) offset`
+  - `i32 (size_t) size`
+  - `i32 (const char*) buffer_data`
+  - `i32 (size_t) buffer_size`
+  - `i32 (uint32_t) flags`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Set content of the buffer `buffer_type` to the bytes (`buffer_data`, `buffer_size`), replacing
+`size` bytes, starting at `offset` in the existing buffer.
```
yuval-k: will this extend the buffer if it is too small?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+Get all key-value pairs from a given map (`map_type`).
+
+
+### `proxy_set_map`
+
+* params:
+  - `i32 (proxy_map_type_t) map_type`
+  - `i32 (const char*) map_data`
+  - `i32 (size_t) map_size`
```
yuval-k: should we document the format of (map_data, map_size) for the get/set ?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+* params:
+  - `i32 (const char*) property_path_data`
+  - `i32 (size_t) property_path_size`
+  - `i32 (const char**) return_property_value_data`
+  - `i32 (size_t*) return_property_value_size`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Get property (`property_path_data`, `property_path_size`).
```
yuval-k: are there a standard set of properties expected to be supported by every host?
the current root id comes to mind

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+
+### `proxy_on_log`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - none
+
+Called in the logging phase.
```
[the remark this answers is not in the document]
htuch: I also commented on https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/2 that I'm a bit skeptical about logging being a first-class phase. Requests end. Extensions can log. Why do we call this explicitly the log phase and not the finalization?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+
+## gRPC callouts
+
+### `proxy_on_grpc_call_response_header_metadata`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (uint32_t) callout_id`
+  - `i32 (size_t) num_elemets`
```
htuch: Nit: `num_elements` (you might want some pedantic spell/format check for this).

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+### `proxy_set_tick_period`
+
+* params:
+  - `i32 (uint32_t) tick_period`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Set timer period (`tick_period`). Once set, the host environment will call `proxy_on_tick` every
+`tick_period` milliseconds.
```
htuch: Would we ever want to consider limiting this call for untrusted code to mitigate timing side-channel attacks?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+  - `i32 (const char*) grpc_service_data`
+  - `i32 (size_t) grpc_service_size`
+  - `i32 (const char*) service_name_data`
+  - `i32 (size_t) service_name_size`
+  - `i32 (const char*) method_data`
+  - `i32 (size_t) method_size`
+  - `i32 (const char*) grpc_message_data`
+  - `i32 (size_t) grpc_message_size`
+  - `i32 (uint32_t) timeout_milliseconds`
+  - `i32 (uint32_t*) return_callout_id`
```
htuch: What about metadata?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+Set timer period (`tick_period`). Once set, the host environment will call `proxy_on_tick` every
+`tick_period` milliseconds.
+
+
+## Buffers, maps, and properties
+
+### `proxy_get_buffer`
+
+* params:
+  - `i32 (proxy_buffer_type_t) buffer_type`
```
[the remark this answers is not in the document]
htuch: I think in general, there are a lot of non-builting types referenced in this doc. They should be given formal definitions in some section.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+  - `i32 (const char*) function_name_data`
+  - `i32 (size_t) function_name_size`
+  - `i32 (const char *) parameters_data`
+  - `i32 (size_t) parameters_size`
+  - `i32 (const char**) return_results_data`
+  - `i32 (size_t*) return_results_size`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Call registered foreign function (`function_name_data`, `function_name_size`).
```
htuch: What is this magical device?

abi-versions/README.md
```diff
@@ -0,0 +1,40 @@
...
+## ABI versioning
+
+The ABI is in the form of: `<major>.<minor>.<patch>`
+
+Breaking changes require increasing the major version.
+
+Adding new functions requires increasing the minor version.
+
+Adding new enum values requires increasing the patch version.
+
```
htuch: One high level question I have is how standardized everything is going to be with WASM. Will we be able to add new callbacks or methods below as the need arises in Envoy extension development, or will this require some external process? Ideally we can review a MVP ABI here and grow it as needed, rather than trying to account for all possible use cases.

PiotrSikora: Ideally, those new callbacks and/or host functions would be generic enough that they would be added to this ABI.

There is a note in the paragraph below about implementation-specific extensions (e.g. `envoy_`), but that's really an escape hatch and hopefully we shouldn't need anything like that in practice.

I fully agree regarding MVP approach. Everything that you see in this ABI was added during Istio Telemetry v2 development, we didn't enumerate all possible cases or add anything that's not already implemented.

abi-versions/README.md
```diff
@@ -0,0 +1,40 @@
...
+Wasm modules should export `proxy_abi_version_<major>_<minor>_<patch>` function, indicating which
+version of the ABI they were compiled against.
+
+Host environments should be able to support multiple versions of the ABI.
```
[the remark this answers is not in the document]
PiotrSikora: This is 100% up to the host implementation, i.e. it can be anywhere between 1 and all.

abi-versions/README.md
PiotrSikora: > * copying ranges of bytes into the WASM address space

`proxy_get_buffer` takes `offset` and `max_size` as a parameter, so this is already supported.

> * searching for bytes/substrings
> * global-replace
> * global-replace-RE2
> * parse as JSON and provide JSON API to content

While those APIs could be implemented, they seem very much like a stretch. We don't have those implemented in `BufferImpl`, for example. Also, it's unclear why those couldn't be implemented inside Wasm module.

> * send content to a socket without brining it into the WASM space

I'm not sure I follow. Where would this content came from? What's the use case?

> I am curious if future WASM VMs or evolutions will support zero copy from outside.

We very much want to support zero-copy, and as soon as Wasm gets support for multiple-memories that can be "mounted" at runtime, we're going to add support for it.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+* params:
+  - none
+* returns:
+  - none
+
+Exports ABI version (vX.Y.Z) this module was compiled against.
+
+Considered alternatives: Ideally, this would be exported as a global variable, but existing
+toolchains have trouble with that (some don’t support global variables at all, some incorrectly
+export language-specific and not Wasm-generic variables).
```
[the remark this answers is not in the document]
PiotrSikora: Correct, the ABI version must be accessible to the host before starting the VM.

abi-versions/vNEXT/README.md
PiotrSikora: You mean returning ID that's allocated by the Wasm VM, that you could (ab)use to return memory address of the context struct? Sure, that's possible, but how many languages would actually benefit from it? Also, it seems that you'd need to keep track of allocated IDs in Wasm VM, to make sure that the `void*` ID in the callback still points to a valid context, so I'm not sure how much performance gain you'd get from this in practice.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+### `proxy_on_context_create`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (uint32_t) parent_context_id`
+* returns:
+  - none
+
+Called when the host environment creates a new root context (if `parent_context_id` is `0`) or a new
+per-stream context.
```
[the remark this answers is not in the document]
PiotrSikora: We use "stream" in network filter as well (since it's operating on a bytestream between downstream and upstream).

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+### `proxy_on_done`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - `i32 (bool) is_done`
+
+Called when the host environment is done processing the context (`context_id`). Return value
+indicates when the Wasm VM is done with the processing as well.
```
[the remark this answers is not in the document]
PiotrSikora: Host will wait until `proxy_done()` is called, I'll add it here.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+### `proxy_on_tick`
+
+* params:
+  - `i32 (uint32_t) root_context_id`
+* returns:
+  - none
+
+Timer called every tick period. Tick period can be configured using
+`proxy_set_tick_period_milliseconds`.
```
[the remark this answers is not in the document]
PiotrSikora: No default, there is no ticking unless configured.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+Called for each chunk of HTTP request body received from the client. Request body can be retrieved
+using `proxy_get_buffer`.
+
+
+### `proxy_on_http_request_trailers`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (size_t) num_trailers`
+  - `i32 (bool) end_of_stream`
```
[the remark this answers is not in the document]
PiotrSikora: I was thinking about HTTP/2 METADATA frames when adding this, but after checking now, it seems that they use `END_METADATA` and not `END_STREAM`. I'll get this fixed.

abi-versions/vNEXT/README.md
PiotrSikora: Erm, sure.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+Note: `downstream` means the connection between client and proxy, and `upstream` means the
+connection between proxy and backend (aka “next hop”).
+
+### `proxy_on_new_connection`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - `i32 (proxy_action_t) next_action`
```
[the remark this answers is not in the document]
PiotrSikora: I'll add this and other enum values shortly.

abi-versions/vNEXT/README.md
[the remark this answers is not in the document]
jplevyak: We could add a token_ptr and if the target token was zero we could use that to indicate that the system should allocate a new timer, but if it was non-zero then it would be the timer to set a new period for.  WDYT?  This would allow multiple timers.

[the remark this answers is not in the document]
mattklein123: +1 for some mechanism to have multiple timers. Also it's not clear to me how to disable a timer so we should document that. I guess set a tick period to 0 for a particular timer token?

abi-versions/vNEXT/README.md
yuval-k: i can see the implementations of most runtimes simplefied by removing the need of a map. essentially every language with a notion of a pointer (c++,rust, assembly-script,tinygo..)

Why would we need keep track of the `void*` ID inside the VM? could it be used with a `proxy_*` callback after `proxy_on_delete`  was called?

abi-versions/vNEXT/README.md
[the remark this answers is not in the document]
jplevyak: yes, that is correct.  we also need to add a uint32_t token to the on_tick call into the VM.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+
+Define a metric using a given name (`metric_name_data`, `metic_name_size`). It can be referred to in
+`proxy_get_metric`, `proxy_increment_metric` and `proxy_record_metric` using returned unique metric
+identifier (`metric_id`).
+
+
+### `proxy_get_metric`
+
+* params:
+  - `i32 (uint32_t) metric_id`
```
jplevyak: In order to read histograms we can add an i32 "value" field which would be the histogram value to read the count from.

abi-versions/vNEXT/README.md
jplevyak: It isn't hard, it is just much less flexible.  If we add another metric type then we have to change the ABI.  Also, the Istio team currently uses another layer which abstracts the different metrics so that they can use a configuration file to define them.  The fact that they can just call proxy_record_metric greatly simplifies the implementation and was a big boon for them.  Given that this is a low level ABI, there is no value IMO in splitting them and great value in keeping it general like it is now.

abi-versions/README.md
jmarantz: proxy_get_buffer sgtm.

The other APIs could help users that want to implement transformations that do not require copying the entire response-body.

The "send content to a socket" could be used for proxying requests to an external transform-server, without the extra copy. One example (which might not be practical as there's already a native impl) is a gzip encoder or jpeg optimizer running out-of-process.

RE wasm support for multiple memories: I don't fully understand what that means, but do you if that's a specific plan from Wasm maintainers?

jmarantz: Also... WDYT about streaming? Say there's a long chunked response; can WASM wake up and process chunks as they come in? Or does the WASM engine not get to look at the first byte until the last byte has arrived?

Apologies if this was already covered.

abi-versions/vNEXT/README.md
mattklein123: From a pure perf perspective I agree with @yuval-k that this would likely work and be usable in many languages. I don't feel strongly if we try to do this now or leave this as a possible post MVP optimization. Perhaps we should start opening issues against this repo?

abi-versions/vNEXT/README.md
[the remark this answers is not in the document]
mattklein123: OK sounds good. You might want to clarify this for non-Envoy people.

abi-versions/vNEXT/README.md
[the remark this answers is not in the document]
mattklein123: +1

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+### `proxy_on_grpc_call_response_message`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (uint32_t) callout_id`
+  - `i32 (size_t) message_size`
+* returns:
+  - none
+
+Called when the response to the gRPC call (`callout_id`) is received.
```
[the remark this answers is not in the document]
mattklein123: Please by overly verbose in the docs for clarity as this is a spec and we should try to make it as easy to understand as possible. Thank you (Similar comment elsewhere.)

abi-versions/vNEXT/README.md
mattklein123: Per the comment in the other PR, if you don't want to merge the function that is fine, but I would probably rename to `lookup` to make it more clear and I would better clarify that `register` is the write side and `lookup` is the read side. cc @jplevyak

abi-versions/vNEXT/README.md
mattklein123: I think we should cover this in our video meeting tomorrow, but briefly, I think it's a false simplification to try to jam all of the different metric types into a single API. It might be possible for counters/gauges (though even that is a bit strange) but it goes totally off the rails for histograms or histogram like things (e.g. the ATS thing you mentioned). I'm not sure how sensible it is to try to get histograms/other into this API.

abi-versions/vNEXT/README.md
mattklein123: The problem is histograms are implementation specific in terms of whether they are HDR, fixed bucket, support summary percentile only, etc. I suspect that for histograms we are going to need a facility to actually read out the buckets, percentiles, etc. I don't see any other sensible way to do this that would support different host types. With that said, I think it's concievable that you can jam other metric types into this API if you squint really hard, but I just don't see how we won't wind up with some histogram specific ABI eventually.

abi-versions/vNEXT/README.md
jplevyak: I disagree.  For histograms you can get a single Y value by providing one additional argument which is the X value.  I don't think we need serialization.  The bucketing should be known through other channels.  It can be in the config for example.

abi-versions/vNEXT/README.md
PiotrSikora: Sure, it's definitely better from the perf point of view, since you're avoiding map lookup for each callback. Having said that, I'm feeling a bit iffy about having pointer to the context maintained outside of the Wasm VM.

How about we allow Wasm VM to return module-assigned ID that would be used in subsequent calls? This way, you could (ab)use this feature and assign the ID corresponding to the memory address. Would that work for you?

abi-versions/vNEXT/README.md
PiotrSikora: Foreign function interface (FFI). Basically, it allows host to register custom functions, that are not part of this ABI, and that will be accessible to the Wasm VM.

abi-versions/vNEXT/README.md
yuval-k: this sounds good to me

abi-versions/README.md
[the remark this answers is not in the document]
htuch: If they only support 1, then they don't need to support multiple versions of the ABI; suggest rewording ^^.

abi-versions/README.md
htuch: I think adding a section calling this out would be helpful. E.g. call it "ABI extension" and describe how FFI is the first expected phase or escape hatch, then the process for having a more generic extension of host functions/callbacks added via WG. What's the process, etc. Which repositories, which PRs, etc.

abi-versions/README.md
kyessenov: I implemented CEL expression foreign function call for this exact reason. It's very expensive to copy data just to do a little computation.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Set property (`property_path_data`, `property_path_size`) to given value (`property_value_data`,
+`property_value_size`).
+
+
+## TCP/UDP/QUIC stream (L4) extensions
+
+### `proxy_resume_downstream`
```
jplevyak: Let's combine these into one proxy_resume(ResumeType) like we have done with the other calls.

In general I think we need enum ResumeType {
  Downstream,
  Upstream,
  Request?
  Response?
}

In the future it is likely that we will want to be able to do the sort of merging of multiple upstream sources into a single downstream flow.   In this case we could extend the enum to include

Stream0 = 100,
Stream99 = 199,

To represent other streams which might be opened by the plugin.

We could also use this for backpressure on say gRPC streams.

abi-versions/vNEXT/README.md
jplevyak: We need to add a note that says that if proxy_on_configure returns false that we will be calling proxy_on_done and then proxy_on_delete before shutting down the VM.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+  - `i32 (size_t) body_size`
+  - `i32 (const char*) trailers_map_data`
+  - `i32 (size_t) trailers_map_size`
+  - `i32 (uint32_t) timeout_milliseconds`
+  - `i32 (uint32_t*) return_callout_id`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Dispatch a HTTP call to upstream (`upstream_name_data`, `upstream_name_size`). Once the response is
+returned to the host, `proxy_on_http_call_response` will be called with a unique call identifier
```
yuval-k, line 670: IIUC, in envoy there is no gaurantee that the http callbacks will not be called in-line with the http call. this is documented for failure case: 

https://github.com/envoyproxy/envoy/blob/95bbc547b8eea39cbab65a58ad49aafd957c14d6/include/envoy/http/async_client.h#L266-L268

but i see filters written defensively and assume that the success callback might be called inline as well. @mattklein123 might know for sure.

if the callbacks are called inline, the calling filter won't know the `callout_id` as it was not returned yet.

two things that come to mind:
- if only the failure callback can be called inline, then it is probably ok, as this is also indicated by the `call_result` return value
- if there is not such guarantee, then perhaps:
  - the calling filter should provide the callout_id ?
  - or perhaps the runtime can guarantee that `proxy_dispatch_http_call` will return before `proxy_on_http_call_response` is called, even if envoy  invokes the callbacks inline?

mattklein123: IIRC it's not possible for the success callback to be called inline (I would need to check) but from an ABI perspective it probably makes sense to make this consistent per @yuval-k suggestions.

abi-versions/vNEXT/README.md
kyessenov: Add note: return_results_data and return_results_size can be nullptr if the function is not expected to return anything.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,919 @@
...
+  - `i32 (proxy_log_level_t) log_level`
+  - `i32 (const char*) message_data`
+  - `i32 (size_t) message_size`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Log message (`message_data`, `message_size`) at the given `log_level`.
+
+
+### `proxy_get_current_time`
```
jplevyak, line 401: We should support __wasi_clock_time_get instead.   This is in keeping with moving toward general WASI function for things which overlap between proxy-wasm and other WASI standards.

abi-versions/vNEXT/README.md
jplevyak: We need an "on_foreign_function" call into the VM, passing a function_id and a data_size.  This will require another BufferType for retrieving the data.

jplevyak:
@PiotrSikora can you give an update?

jplevyak:
With no action in 4 months, this PR is abandoned.   Closing it.

PiotrSikora:
I believe most (all?) comments were addressed, but some things might have been lost in the rewrite, so please take a look. Thanks!

PiotrSikora:
@mattklein123 @htuch @yskopets @yuval-k @gbrail @jplevyak @kyessenov could you please review the latest iteration of the ABI spec? Thanks!

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (uint32_t) parent_context_id`
+  - `i32 (proxy_context_type_t) context_type`
+* returns:
+  - `i32 (uint32_t) new_context_id`
+
+Called when the host environment creates a new context (`context_id`) of type
+`context_type` (e.g. `VmContext`, `PluginContext`, `StreamContext`,
```
yskopets, line 80: Is this spec supposed to cover `Access Logger` extensions ?

Do `Access Logger` extensions have a dedicated context type ?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+Allocates continuous memory buffer of size `memory_size` using the in-VM memory
+allocator and returns address of the memory buffer to the host environment using
+`memory_data`. Returning `0` indicates failure.
+
+
+## Context lifecycle
+
+### Callbacks optionally exposed by the Wasm module
+
+#### `proxy_on_context_create`
```
yskopets, line 70: Is it possible to use `proxy_on_context_create ` callback to achieve a behaviour where an extension can have multiple instances within a VM each with a different configuration ?

E.g.,
* 1 `Envoy` `Listener` with `HttpConnectionManager` and 2 instance of the same `Http Filter` extension but different configs
* 2 different `Envoy` `Listener`s, each with `HttpConnectionManager` and 1 instance of the same `Http Filter` extension but different configs

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  - `i32 (uint32_t) callout_id`
+  - `i32 (uint32_t) status_code`
+* returns:
+  - none
+
+Called when the request to the gRPC call (`callout_id`) fails or if the gRPC
+connection is closed.
+
+
+## Custom extension points
```
yskopets, line 1315: Would it possible to describe example use case ?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+### Callbacks optionally exposed by the Wasm module
+
+#### `proxy_on_custom_callback`
```
yskopets, line 1319: Is it possible to call back into Plugin context / Http Stream context / TCP stream context ?

How can extension understand for what context it is intended ?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+Note: `downstream` means the connection between client and proxy, and `upstream`
+       means the connection between proxy and backend (aka “next hop”).
+
+Note: The same unique context identifier (`stream_id`) is shared by a pair of
+      `downstream` and `upstream` connections.
+
+### Callbacks optionally exposed by the Wasm module
+
+#### `proxy_on_new_connection`
```
kyessenov, line 184: Can you specify the state machine of the callbacks? E.g. 
```regex
new_connection (downstream_data | upstream_data)* (downstream_close upstream_data* upstream_close | upstream_close downstream_data* downstream_close)
```

Also, specify the interleavings with configuration callbacks.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+This callback can be used to associate `context_id` with `parent_context_id`,
+or when using custom context identifiers, but otherwise the in-VM context can be
+implicitly created in `proxy_on_vm_start`, `proxy_on_plugin_start`,
+`proxy_on_new_connection` and `proxy_on_http_request_headers` callbacks.
+
+Return value (`new_context_id`) can be used to assign new context identifier
+that will be used by the host environment in subsequent calls. Returning `0`
+retains the original context identifier.
+
+#### `proxy_on_context_finalize`
```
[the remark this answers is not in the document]
kyessenov: Right, where is `on_log`?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+#### `proxy_context_finalize`
+
+* params:
+  - none
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Indicates to the host environment that the Wasm module finished processing
+current context. It should be used after previously returning `false` in
```
gbrail, line 131: "Must" be used ... otherwise the module creates a memory leak, right?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that data chunk should be forwarded upstream.
+- `EndStream` means that data chunk should be forwarded upstream and that this
+  connection should be closed afterwards.
+- `Done` means that data chunk should be forwarded upstream and that this
+  extension shouldn't be called again for the current stream.
+- `Pause` means that further processing should be paused until
+  `proxy_resume_stream` is called.
+- `WaitForMoreData` means that further processing should be paused and that host
```
gbrail, line 222: Is there a way for a network filter to short-circuit processing by writing some data back downstream, like "send_local_response" for HTTP filters? Imagine a filter that reads some authorization information from a connection and, if invalid, sends back an error message.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+Headers (`num_headers` entries) can be retrieved using `proxy_get_map_values`.
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that this HTTP request should be accepted and that HTTP
+  headers should be forwarded upstream.
+- `EndStream` means that this HTTP request should be accepted, that HTTP headers
+  should be forwarded upstream and that the underlying stream should be closed
+  afterwards.
+- `Done` means that this HTTP request should be accepted, that HTTP headers
```
gbrail, line 338: In this case I assume that "finalize" is still called?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+Called when HTTP request headers for stream `stream_id` are received from
+downstream.
+
+Headers (`num_headers` entries) can be retrieved using `proxy_get_map_values`.
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that this HTTP request should be accepted and that HTTP
+  headers should be forwarded upstream.
+- `EndStream` means that this HTTP request should be accepted, that HTTP headers
```
gbrail, line 335: Does "afterwards" mean "after the headers are sent," or some other time? I can't say I understand the usefulness of this one.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+* params:
+  - `i32 (uint32_t) stream_id`
+  - `i32 (size_t) num_trailers`
+  - `i32 (bool) end_of_stream`
+* returns:
+  - `i32 (proxy_action_t) next_action`
+
+Called when HTTP request trailers for stream `stream_id` are received from
+downstream.
```
gbrail, line 385: If there are no trailers, is this callback still invoked? It might be hard to handle errors in a protocol that is expecting trailers and then doesn't get them.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+Sends HTTP response to the client, without forwarding request to the upstream.
+
+Returns `Ok` on success.
+
+Returns `BadArgument` for invalid `response_code` and/or `grpc_status`.
+
+Returns `InvalidMemoryAccess` when `response_code_details_data`,
+`response_code_details_size`, `response_body_data`, `response_body_size`,
+`additional_headers_map_data` and/or `additional_headers_size` point to invalid
+memory address.
```
gbrail, line 548: I recall from Envoy that it's only valid to call this on the request path -- on the response path strange things happen. Can we clear that up in the docs and return an error if the module calls it at the wrong time?

Also, once this is called, are other callbacks, like proxy_on_response headers and data, called on the response? Or is everything skipped now?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+#### `proxy_abi_version_X_Y_Z`
+
+* params:
+  - none
+* returns:
+  - none
+
+Exports ABI version (vX.Y.Z) that the Wasm module was compiled against.
+
+#### `proxy_log`
```
gbrail, line 34: Can we also implement "proxy_get_log_level?" It's handy to initialize the logging library with the caveat that one shouldn't call it on every long message and that we haven't figured out how to call back when the log level changes.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+buffer if target `size` is larger than `buffer_size`).
+
+Returns `Ok` on success.
+
+Returns `BadArgument` for unknown `buffer_type`.
+
+Returns `InvalidMemoryAccess` when `buffer_data` and/or `buffer_size` point to
+invalid memory address.
+
+#### `proxy_get_map_values`
```
gbrail, line 635: What happened to proxy_(get,set,replace)_header_map_value?

It's very common IMO for an HTTP module to only manipulate a few headers. It might be more efficient, and it's certainly less complicated, to access or change those directly via the callback versus getting the whole map and then having to replace (correctly) the whole map.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+### Functions exposed by the host environment
+
+#### `proxy_open_shared_queue`
+
+* params:
+  - `i32 (const char*) queue_name_data`
+  - `i32 (size_t) queue_name_size`
+  - `i32 (bool) create_if_not_exist`
+  - `i32 (uint32_t*) return_queue_id`
```
gbrail, line 863: Are "queue_ids" intended to be unique across all wasm modules on the same host, or only on the same vm, or only on the same module?

I'm asking because in one case I found it useful for a singleton "bootstrap service" to expose a "request queue," and for each instance of another module to create its own "reply queue." The non-singleton module then puts a message on the request queue that includes the queue id of the reply queue, so that the two can implement a request-response workflow.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  - `i32 (uint32_t) queue_id`
+* returns:
+  - none
+
+Called when new item is available in the queue (`queue_id`).
+
+
+## Timers
+
+### Functions exposed by the host environment
```
gbrail, line 948: What happened to the method to get the current timestamp? Without it, then we need to document that the host must expose some sort of WASI interface.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+Once the response is returned to the host, `proxy_on_http_call_response` will be
+called with an unique call identifier (`return_callout_id`).
+
+Returns `Ok` on success.
+
+Returns `NotFound` for unknown upstream.
+
+Returns `NotAllowed` when requests to given upstream are not allowed.
+
+Returns `InvalidMemoryAccess` when `upstream_name_data`, `upstream_name_size`,
```
gbrail, line 1116: FWIW I learned that Envoy will return an error in this call if the ":path", ":method", and "host" headers aren't set. In general we should say something about what the various header fields mean to the host. Do we want to document that?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+
+## gRPC callouts
+
+### Functions exposed by the host environment
+
+#### `proxy_dispatch_grpc_call`
+
+* params:
+  - `i32 (const char*) upstream_name_data`
```
gbrail, line 1148: So in this version can this just be the name of the upstream? As discussed elsewhere this is currently a gnarly protobuf.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+Once the response is returned to the host, `proxy_on_grpc_call_response` and/or
+`proxy_on_grpc_call_close` will be called with an unique call identifier
+(`return_callout_id`). The unique call identifier can also be used to cancel
+outstanding requests using `proxy_cancel_grpc_call` or close outstanding request
+using or `proxy_close_grpc_call`.
+
+Returns `Ok` on success.
+
+Returns `NotFound` for unknown service (`service_name_data`,
+`service_name_size`).
```
gbrail, line 1175: Doesn't that error come back from the upstream gRPC server? In that case would the error really be there or in the response? OTOH, I imagine that an invalid upstream name would immediately return an error.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  - `i32 (size_t) grpc_message_size`
+  - `i32 (uint32_t) timeout_milliseconds`
+  - `i32 (uint32_t*) return_callout_id`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Dispatch a call to a gRPC service (`service_name_data`, `service_name_size`)
+using upstream (`upstream_name_data`, `upstream_name_size`).
+
+Once the response is returned to the host, `proxy_on_grpc_call_response` and/or
```
gbrail, line 1166: Do we mean proxy_on_grpc_call_response (which I can't find) or something else?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+* params:
+  - `i32 (proxy_map_type_t) map_type`
+  - `i32 (iovec_array_t) keys_data`
+  - `i32 (size_t) keys_size`
+  - `i32 (iovec_array_t[]*) return_map_data`
+  - `i32 (size_t*) return_map_size`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Returns map containing key-values for keys from the provided list of vectors
```
gbrail, line 646: If this map is going to be used for HTTP headers, then do we want to standardize what names are there? Envoy has a particular set of system-defined headers like :path, :method, etc, and the rest are just lower-cased versions of the HTTP header. Should the spec discuss this so that people can write modules that can theoretically support multiple hosts?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+All functions exposed by the host environment return `proxy_result_t`, which
+indicates the status of the call (success, invalid memory access, etc.), and
+return values are written into memory pointed by `return_<value>` parameters.
+
+
+## Integration
+
+### Functions exposed by the host environment
+
+#### `proxy_abi_version_X_Y_Z`
```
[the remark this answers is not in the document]
PiotrSikora: Thanks!

abi-versions/vNEXT/README.md
PiotrSikora: Yeah, I'll add both.

abi-versions/vNEXT/README.md
PiotrSikora: We don't have that right now. I can add `proxy_send_bytes_downstream` if you want.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+- `Continue` means that this HTTP request should be accepted and that HTTP
+  headers should be forwarded upstream.
+- `EndStream` means that this HTTP request should be accepted, that HTTP headers
+  should be forwarded upstream and that the underlying stream should be closed
+  afterwards.
+- `Done` means that this HTTP request should be accepted, that HTTP headers
+  should be forwarded upstream and that this extension shouldn't be called again
+  for this HTTP request.
+- `Pause` instructs the host environment to pause processing until
+  `proxy_resume` is called.
```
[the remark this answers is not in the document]
PiotrSikora: This should be `proxy_http_resume`.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+#### `proxy_dequeue_shared_queue_item`
+
+* params:
+  - `i32 (uint32_t) queue_id`
+  - `i32 (iovec_array_t*) return_payload_data`
+  - `i32 (size_t*) return_payload_size`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Removes item from the tail of the queue (`queue_id`). Returned value is written
```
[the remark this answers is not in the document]
PiotrSikora: Oops, it looks that I accidentally removed it during rewrite. It should return `Empty`.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+Deletes previously created shared queue (`queue_id`).
+
+Returns `Ok` on success.
+
+Returns `BadArgument` when the queue (`queue_id`) doesn't exist.
+
+### Callbacks optionally exposed by the Wasm module
+
+#### `proxy_on_queue_ready`
```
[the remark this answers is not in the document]
PiotrSikora: It's not correlated with any `context_id`, since queue is a standalone object.

`proxy_open_shared_queue` returns unique `queue_id` handle, which is used in `proxy_on_queue_ready`.

Always notifying all WasmVms seems wasteful, so we should do some kind of load-balancing (probably with ACK/NACK), but I guess we could declare each queue as either unicast or multicast upon creation, if you have a use case for it?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+Increment/decrement value of the metric (`metric_id`) by offset (`offset`).
+
+Returns `Ok` on success.
+
+Returns `BadArgument` when the metric (`metric_id`) doesn't exist.
+
+Returns `InvalidOperation` when `metric_type` doesn't support incrementing
+and/or decrementing values.
+
+#### `proxy_delete_metric`
```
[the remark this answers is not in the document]
PiotrSikora: In theory, it's here to remove metrics from the host and avoid resource leaks, but I'm not sure how useful this will be in practice. It probably should match number of `proxy_create_metric` calls, but I could be convinced otherwise.

yuval-k:
i don't have any further comments.
thank you for making this happen!

abi-versions/vNEXT/README.md
yskopets: > No, since logging can happen in the "finalize" calls.

Handling access logs typically involves buffering, timeout, http/grpc call. All these actions have less sense in the context of an individual HTTP request / TCP connection than in the context of a dedicated `Access Logger` extension.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+retrieved using `proxy_get_map_values`.
+
+Response body (of size `body_size`) can be retrieved using `proxy_get_buffer`.
+
+
+## gRPC callouts
+
+### Functions exposed by the host environment
+
+#### `proxy_dispatch_grpc_call`
```
htuch, line 1145: Why is there a difference in the ABI between a call and a stream? Why can't we just treat everything like a stream and let higher level code deal with the distinction?

htuch: The only difference I think is in terms of ideas like timeout, which the SDK can potentially handle itself.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+`values_size` point to invalid memory address.
+
+
+## Shared key-value store
+
+Note: Global key-value store can be always accessed using `kvstore_id` `0`.
+
+### Functions exposed by the host environment
+
+#### `proxy_open_shared_kvstore`
```
htuch, line 701: Why can't the kvstore be used to model headers? It seems a bit awkward to have two different ABI mechanisms for key-values.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+Called on a new connection (`stream_id`).
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that this connection should be accepted.
+- `Done` means that this connection should be accepted and that this extension
+  shouldn't be called again for the current stream.
+- `Pause` means that further processing should be paused until
+  `proxy_resume_stream` is called.
+- `Close` means that this connection should be rejected.
```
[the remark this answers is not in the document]
htuch: Actually, I see why this applies for `proxy_on_new_connection`, but not for `proxy_on_downstream_data`.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+#### `proxy_on_http_response_headers`
+
+* params:
+  - `i32 (uint32_t) stream_id`
+  - `i32 (size_t) num_headers`
+  - `i32 (bool) end_of_stream`
+* returns:
+  - `i32 (proxy_action_t) next_action`
+
+Called when HTTP response headers for stream `stream_id` are received from
```
htuch, line 431: Are you planning on explicitly modeling 1xx in the ABI?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+Returns `InvalidOperation` when stream wasn't paused.
+
+#### `proxy_close_stream`
+
+* params:
+  - `i32 (proxy_stream_type_t) stream_type`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Close downstream or upstream (`stream_type`) connections in current context.
```
alyssawilk, line 302: see comments above about Envoy close types

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+#### `proxy_on_http_request_body`
+
+* params:
+  - `i32 (uint32_t) stream_id`
+  - `i32 (size_t) body_size`
+  - `i32 (bool) end_of_stream`
+* returns:
+  - `i32 (proxy_action_t) next_action`
+
+Called for each chunk of HTTP request body for stream `stream_id` received from
```
alyssawilk, line 354: chunk in the context of HTTP could be confused for HTTP/1.1  chunks.  rephrase or clarify?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  for this HTTP request.
+- `Pause` instructs the host environment to pause processing until
+  `proxy_resume` is called.
+- `Close` means that the stream should be closed immediately.
+
+#### `proxy_on_http_request_body`
+
+* params:
+  - `i32 (uint32_t) stream_id`
+  - `i32 (size_t) body_size`
```
alyssawilk, line 349: body size or size of this chunk?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+upstream.
+
+Trailers (`num_trailers` entries) can be retrieved using `proxy_get_map_values`.
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that HTTP trailers should be forwarded downstream.
+- `EndStream` means that HTTP trailers should be forwarded downstream and that
+  the underlying stream should be closed afterwards.
+- `Done` means that HTTP trailers should be forwarded downstream and that this
+  extension shouldn't be called again for this HTTP response.
```
alyssawilk, line 495: Are EndStream and Done redundant for trailers calls?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+Request body (of size `body_size`) can be retrieved using `proxy_get_buffer`.
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that this HTTP request body chunk should be forwarded
+  upstream.
+- `EndStream` means that this HTTP request body chunk should be forwarded
+  upstream and that the underlying stream should be closed afterwards.
+- `Done` means that this HTTP request body should be forwarded upstream and that
+  this extension shouldn't be called again for this HTTP request.
```
alyssawilk, line 365: I think for HTTP bodies specifically it would be useful to have DoneWithData.

I could easily imagine filters which don't want to process body data (or incur the context switch to even be informed about it) but do want to interact with trailers (and maybe metadata).

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+- `Pause` means that further processing should be paused until `proxy_resume` is
```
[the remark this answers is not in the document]
alyssawilk: WaitForMoreData would call this callback when more data arrives where Pause would swallow that event until the wasm filter resumed.

I assume this maps to StopAllIterationAndBuffer, rather than StopIteration (which is kind of the Envoy default) or StopAllItreationAndWatermark.  Might be worth clarifying?

abi-versions/vNEXT/README.md
alyssawilk: I don't think we've added the API yet but it's on my todo list to allow trySendLocalReply on the encode path (see comments in sendLocalReply for the 3 best-effort modes)

given wasm APIs are supposed to be stable it's worth considering starting with this from the get go.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+Logs message (`message_data`, `message_size`) at the log level (`log_level`).
+
+Returns `Ok` on success.
+
+Returns `BadArgument` for unknown `log_level`.
+
+Returns `InvalidMemoryAccess` when `message_data` and/or `message_size` point to
+invalid memory address.
+
+### Callbacks optionally exposed by the Wasm module
```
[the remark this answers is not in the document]
mattklein123: Yeah I would have a required section and be clear that a module will fail to load unless it implements all required exports.

abi-versions/vNEXT/README.md
mattklein123, line 131: nit: you might want to mention `proxy_set_effective_context` which I assume might need to be called if the contexts have switched before finalizing.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+* returns:
+  - `i32 (proxy_action_t) next_action`
+
+Called on a new connection (`stream_id`).
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that this connection should be accepted.
+- `Done` means that this connection should be accepted and that this extension
+  shouldn't be called again for the current stream.
+- `Pause` means that further processing should be paused until
```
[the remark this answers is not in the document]
mattklein123: Last we talked about this the API was going to replicate all of the filter return codes we already have, so watermarking should be included, though I haven't read that far yet.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+* params:
+  - `i32 (uint32_t) vm_id`
+  - `i32 (size_t) vm_configuration_size`
+* returns:
+  - `i32 (bool) is_success`
+
+Called when the host environment starts WebAssembly Virtual Machine (WasmVm).
+
+Its configuration (of size `vm_configuration_size`) can be retrieved using
+`proxy_get_buffer`.
```
mattklein123, line 152: FWIW I find this proxy_get_buffer dance to be a bit magical from an API design perspective. What if in the future we have two buffers that we want to get? IMO for this type of thing I think it would be much cleaner and more future proof to pass an opaque_handle that could then be passed back to proxy_get_buffer, or a size and handle if that makes more sense. This would make it very explicit what the user is fetching if we ever need to add the ability to fetch multiple things on the context of an API call.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  - `i32 (proxy_action_t) next_action`
+
+Called for each data chunk received from downstream (`stream_id`).
+
+Data (of size `data_size`) can be retrieved using `proxy_get_buffer`.
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that data chunk should be forwarded upstream.
+- `EndStream` means that data chunk should be forwarded upstream and that this
+  connection should be closed afterwards.
```
mattklein123, line 217: What does this mean in terms of half duplex? It's a bit odd to me that this is a return code. Would it be better to have this happen via a callback on the connection?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+- `Done` means that data chunk should be forwarded upstream and that this
+  extension shouldn't be called again for the current stream.
+- `Pause` means that further processing should be paused until
```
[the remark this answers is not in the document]
mattklein123: Also would be good to comment on watermarking/buffering as part of this.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  `proxy_resume_stream` is called.
+- `WaitForMoreData` means that further processing should be paused and that host
+  should call this callback again when it has more data.
+- `WaitForEndOrFull` means that further processing should be paused and that
+  host should call this callback again when downstream connection is closed or
+  when its internal buffers are full.
+- `Close` means that this connection should be closed immediately.
```
mattklein123, line 227: Same question about half duplex.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  connection should be closed afterwards.
+- `Done` means that data chunk should be forwarded upstream and that this
+  extension shouldn't be called again for the current stream.
+- `Pause` means that further processing should be paused until
+  `proxy_resume_stream` is called.
+- `WaitForMoreData` means that further processing should be paused and that host
+  should call this callback again when it has more data.
+- `WaitForEndOrFull` means that further processing should be paused and that
+  host should call this callback again when downstream connection is closed or
+  when its internal buffers are full.
```
mattklein123, line 226: What does this mean? What buffers? How does this relate to watermarking?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  - `i32 (proxy_action_t) next_action`
+
+Called for each data chunk received from upstream (`stream_id`).
+
+Data (of size `data_size`) can be retrieved using `proxy_get_buffer`.
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that data should be forwarded downstream.
+- `EndStream` means that data should be forwarded downstream and that this
+  connection should be closed afterwards.
```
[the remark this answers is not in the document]
mattklein123: I commented on this above. It's not super clear to me that closing should be a return code from the filter vs. a callback API, but I agree we need to be able to do all of these types.

abi-versions/vNEXT/README.md
mattklein123: And if we have this why do we need the return codes for end stream, etc.? Can we just have a stop iteration return code?

abi-versions/vNEXT/README.md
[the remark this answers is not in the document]
mattklein123: In general it seems like we are missing the options around watermarking vs. buffering. I think that is needed. I commented on this elsewhere.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+
+Called when HTTP request trailers for stream `stream_id` are received from
+downstream.
+
+Trailers (`num_trailers` entries) can be retrieved using `proxy_get_map_values`.
+
+Return value (`next_action`) instructs the host environment what action to take:
+- `Continue` means that HTTP trailers should be forwarded upstream.
+- `EndStream` means that HTTP trailers should be forwarded upstream and that
+  the underlying stream should be closed afterwards.
```
mattklein123, line 392: How is this not the same as Continue? I think this can be removed as it's implicit?

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+- `Done` means that HTTP trailers should be forwarded upstream and that this
+  extension shouldn't be called again for this HTTP request.
```
mattklein123, line 394: Not possible I don't think

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+Resume processing of paused HTTP request or response (`stream_type`) in current
+context.
+
+Returns `Ok` on success.
+
+Returns `BadArgument` for unknown `stream_type`.
+
+Returns `InvalidOperation` when stream wasn't paused.
+
+#### `proxy_close_http_stream`
```
mattklein123, line 566: Similar comment about if we have this why do we need the special return codes from the filter callbacks. Seems like it would be simpler to reduce the return codes there and just have people call this before they stop iteration or whatever, though I don't feel strongly about it.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  - `i32 (size_t) max_size`
+  - `i32 (iovec_array_t*) return_buffer_data`
+  - `i32 (size_t*) return_buffer_size`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Returns content of the buffer `buffer_type`. At most `max_size` bytes will be
+returned, starting from the `offset`. Returned bytes are written into memory
+slices pointed by a list of vectors (`return_buffer_data`,
+`return_buffer_size`).
```
mattklein123, line 600: Can you clarify this? The memory is allocated by callout via the allocate API, right? I'm not completely clear on what are in/out params here and how this works. I think more text would help.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+  - `i32 (offset_t) offset`
+  - `i32 (size_t) size`
+  - `i32 (iovec_array_t) buffer_data`
+  - `i32 (size_t) buffer_size`
+* returns:
+  - `i32 (proxy_result_t) call_result`
+
+Sets content of the buffer `buffer_type` to the values from memory slices
+pointed by the provided list of vectors (`buffer_data`, `buffer_size`),
+replacing `size` bytes in the existing buffer, starting at the `offset`. Host
```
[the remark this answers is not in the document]
mattklein123: Can you add this type of thing into the spec? It would help similar to my comments above.

abi-versions/vNEXT/README.md
```diff
@@ -0,0 +1,1471 @@
...
+* params:
+  - `i32 (uint32_t) timer_id`
+* returns:
+  - none
+
+Called after configured period by the timer (`timer_id`) created using
+`proxy_create_timer`.
+
+
+## Stats/metrics
```
mattklein123, line 996: Where did we land on the differences between counters/gauges/histograms? Do we need a histogram API? It's not clear to be from reading the spec how a generalized API makes sense, or it seems like it should be more clear about what APIs are supported on what types and otherwise will return an error.

abi-versions/vNEXT/README.md
mattklein123: +1 not totally clear how this would be used. An example would be great.

abi-versions/vNEXT/README.md
htuch: +1. I was going to comment on this, but I realized that for the practical use cases in the existing ABI that the key/value method seemed to work fine. But, I suspect that the proposal here will future proof and make the ABI a little more regular.

seblaz:
Hi @PiotrSikora, is there something I can do to help you with this changes?

PR #5, opened 2020-04-17
Add spec for a Tracing ABI

mohit-a21:
Added proxy_set_active_span_tag to the abi. Using this ABI,
we should be able to add tag key and value to an active span
in case tracing is enabled for a request.


mohit-a21:
@PiotrSikora Please review this.

jcchavezs:
I have collected some thoughts on this in https://github.com/corazawaf/coraza-proxy-wasm/issues/166 and https://github.com/envoyproxy/envoy/issues/21959

PR #39, opened 2023-05-08
rename proxy_set_tick_period to reflect resolution

bonifaido:
this function is named proxy_set_tick_period_milliseconds at other parts of the spec and across implementations


mathetake:
Hi @bonifaido, please refer to the context https://github.com/proxy-wasm/spec/pull/40 but tldr is that the vNext/README.md has never been correct. We maintainers are gearing toward project overhaul, so let me close this PR for now. Thank you for your understanding!

PR #38, opened 2023-03-13
Add proxy_log_destination ABI

vikaschoudhary16:
This new ABI will help in usecase where it is intended to separate logs from the plugins based on the domain for example audit logs.
xref: https://github.com/envoyproxy/envoy/issues/22669

Issue #36, opened 2022-11-22
Inconsistency between the spec and implementation. 


butterunderflow:
Hello, I'm a newcomer to proxy-wasm.

It seems there has been some inconsistency between the spec and implementation.
For example, the host interface `proxy_get_map_value` is not used in [Go SDK](https://github.com/tetratelabs/proxy-wasm-go-sdk) nor [Rust SDK](https://github.com/proxy-wasm/proxy-wasm-rust-sdk), they use  `proxy_get_header_map_value` instead. [For example](https://github.com/tetratelabs/proxy-wasm-go-sdk/blob/fcfdccc0150019ff006cc5ff3c932dbe2b5ec080/proxywasm/internal/abi_hostcalls.go#L44). 

Is this intentional or is it a choice of implementations?

codefromthecrypt:
ran into another one today: proxy_send_http_response here, but proxy_send_local_response in the impls. For example, https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/d32cb05cb666216db5e1a383c1ffa2c1aeec19bb/include/proxy-wasm/wasm_api_impl.h#L98

mathetake:
I am closing this issue, but please refer to https://github.com/proxy-wasm/spec/pull/39#issuecomment-1577876577 for the context!

PR #40, opened 2023-06-06
Deletes unimplemented vNEXT doc.

mathetake:
As a part of the project overhaul following the discussion in https://github.com/proxy-wasm/spec/pull/38#issuecomment-1577831426, 
this commit deletes the never-implemented vNEXT ABI documentation 
which has caused confusion to users and those willing to contribute.

The latest ["implemented" specification version is `v0.2.1`](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/src/wasm.cc#L128-L131), and the proper 
documentation of v0.2.1 will be added to this repository as a follow-up PR.


mathetake:
cc @mpwarres @martijneken @vikaschoudhary16 @jcchavezs

jcchavezs:
Good to see some movement here!

mpwarres (approved):
LGTM. Would like to get @PiotrSikora 's ok on this as well.

jcchavezs:
Ping @PiotrSikora

mathetake:
I believe the agreement is already made between @PiotrSikora and @mpwarres as-in https://github.com/proxy-wasm/spec/pull/38#issuecomment-1577831426, so I will go ahead and merge this. Thanks!

PiotrSikora:
This is wrong and shouldn't have been merged.

> I believe the agreement is already made between @PiotrSikora and @mpwarres as-in [#38 (comment)](https://github.com/proxy-wasm/spec/pull/38#issuecomment-1577831426), so I will go ahead and merge this. Thanks!

@mpwarres clearly asked for my approval on this PR in https://github.com/proxy-wasm/spec/pull/40#pullrequestreview-1468931815, so I don't understand why you chose to "believe" otherwise.

codefromthecrypt:
not sure @mathetake [replied](https://github.com/proxy-wasm/spec/issues/41#issuecomment-1588235039) to the correct issue on vNEXT or not. 

I routinely point people to the [issue that says it is broken](https://github.com/proxy-wasm/spec/issues/36). No one is looking to cause any problems, but we do want to avoid pain. Not two weeks ago, Takeshi and I had to privately explain why this document is confusing to an employee at a company we don't work at (redacted to protect the innocent)

Screenshot 2023-06-13 at 07 57 24

I'm not trying to convince you that deleting is the right answer, but it certainly wasn't motivated for any reason except to stop the constant damage to our time.

All this said, usually when I'm in an OSS project and someone with access merged something I don't agree with.. If I feel strongly, I'll revert it. I think if you want to revert it, even if that position is unpopular, others would accept it.

So, TL;DR; is my unsolicited 2p as a consumer not a committer, is revert if you feel this was wrong!

PiotrSikora:
> I routinely point people to the [issue that says it is broken](https://github.com/proxy-wasm/spec/issues/36). No one is looking to cause any problems, but we do want to avoid pain. Not two weeks ago, Takeshi and I had to privately explain why this document is confusing to an employee at a company we don't work at (redacted to protect the innocent)
> 
> Screenshot 2023-06-13 at 07 57 24
> I'm not trying to convince you that deleting is the right answer, but it certainly wasn't motivated for any reason except to stop the constant damage to our time.

It says `vNEXT` right there in the link... but why are they looking at the ABI spec in the first place? I feel like there is some context missing.

PR #42, opened 2023-07-04
Document Proxy-Wasm ABI v0.2.1.

### LINKED ISSUE #41: Properly document v0.2.1 ABI
Unfortunately, Proxy-Wasm ABI hasn't been appropriately documented since the beginning and has started being used by Envoy and Istio without the proper reference except for the code base of Envoy and Proxy-Wasm SDK/host implementations. That has caused a lot of confusion to anyone interested in this project, not only the end users but also those who is willing to contirbute.

As the former/current leads of this project discussed in the comments https://github.com/proxy-wasm/spec/pull/38#issuecomment-1575749431 to break this standstill situation which has lasted for the last three years, we agreed on correctly documenting the currently implemented v0.2.1 ABI which is the de-factor in this space (implemented Envoy and other proxies).

After this is resolved, we can incrementally discuss and fix the issues associated with the current ABI, such as #5 #32 #38.

cc @vikaschoudhary16  @jcchavezs @anuraaga @PiotrSikora @martijneken @mpwarres @codefromthecrypt @M4tteoP

PiotrSikora:
Fixes #41.

### DIFF
--- a/abi-versions/v0.2.1/README.md
+++ b/abi-versions/v0.2.1/README.md
@@ -898,6 +898,9 @@ Plugin must return one of the following values:
 
 Called when downstream connection is closed.
 
+The `peer_type` should describe whether this was initiated by a `LOCAL`
+or `REMOTE` peer, but this value might also be `UNKNOWN`.
+
 
 #### `proxy_on_upstream_data`
 
@@ -937,6 +940,9 @@ Plugin must return one of the following values:
   - none
 
 Called when upstream connection is closed.
+
+The `peer_type` should describe whether this was initiated by a `LOCAL`
+or `REMOTE` peer, but this value might also be `UNKNOWN`.
 
 
 ## HTTP streams
--- a/abi-versions/v0.2.1/README.md
+++ b/abi-versions/v0.2.1/README.md
@@ -411,6 +411,8 @@ Returned `errno` value is:
 Sets timer period (`tick_period`). When set, the host will call
 [`proxy_on_tick`] every `tick_period` milliseconds.
 
+Setting `tick_period` to `0` disables the timer.
+
 Returned `status` value is:
 - `OK` on success.
 
--- a/abi-versions/v0.2.1/README.md
+++ b/abi-versions/v0.2.1/README.md
@@ -1,0 +1,1996 @@
+# Proxy-Wasm ABI v0.2.1 specification
+
+---
+
+Status: **FINAL**
+
+---
+
+# Callbacks and Functions
+
+### Callbacks exposed by the Wasm module
+
+All callbacks (entry points named `proxy_on_<event>`) are optional,
+and are going to be called only if exposed by the Wasm module.
+
+All callbacks, other than the [integration] and [memory management]
+functions, include a unique context identifier as the first parameter,
+which can be used to distinguish between different contexts.
+
+
+### Functions exposed by the host
+
+All functions exposed by the host are required.
+
+All Proxy-Wasm functions exposed by the host return [`proxy_status_t`],
+which indicates status of the call (success, invalid memory access,
+etc.). Return values are written into memory pointed by `return_<value>`
+parameters.
+
+
+## Integration
+
+### Callbacks exposed by the Wasm module
+
+#### `proxy_abi_version_0_2_1`
+
+* params:
+  - none
+* returns:
+  - none
+
+Function marker used during linking to advertise Wasm module's support
+for Proxy-Wasm ABI v0.2.1.
+
+This function is never called.
+
+
+#### `_initialize`
+
+* params:
+  - none
+* returns:
+  - none
+
+Called when the Wasm module is first loaded.
+
+
+#### `main`
+
+* params:
+  - `i32 (uint32_t) ignored`
+  - `i32 (uint32_t) ignored`
+* returns:
+  - `i32 (uint32_t) ignored`
+
+> **Warning**
+> This is called only if [`_initialize`] is also exported.
+
+Called when the Wasm module is first loaded, after [`_initialize`].
+
+
+#### `_start`
+
+* params:
+  - none
+* returns:
+  - none
+
+> **Warning**
+> This is called only if [`_initialize`] is not exported.
+
+Called when the Wasm module is first loaded.
+
+
+## Memory management
+
+### Callbacks exposed by the Wasm module
+
+#### `proxy_on_memory_allocate`
+
+* params:
+  - `i32 (size_t) memory_size`
+* returns:
+  - `i32 (uint8_t *) memory_data`
+
+Called to allocate continuous memory buffer of `memory_size` using
+the in-VM memory allocator.
+
+Plugin must return `memory_data` pointing to the start of the allocated
+memory.
+
+Returning `0` indicates failure.
+
+
+#### `malloc`
+
+* params:
+  - `i32 (size_t) memory_size`
+* returns:
+  - `i32 (uint8_t *) memory_data`
+
+> **Warning**
+> This callback has been deprecated in favor of [`proxy_on_memory_allocate`],
+> and it's called only in its absence.
+
+Called to allocate continuous memory buffer of `memory_size` using
+the in-VM memory allocator.
+
+Plugin must return `memory_data` pointing to the start of the allocated
+memory.
+
+Returning `0` indicates failure.
+
+
+## Context lifecycle
+
+### Callbacks exposed by the Wasm module
+
+#### `proxy_on_context_create`
+
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (uint32_t) parent_context_id`
+* returns:
+  - none
+
+Called when the host creates a new context (`context_id`).
+
+When `parent_context_id` is `0` then a new plugin context is created,
+otherwise a new per-stream context is created.
+
+
+#### `proxy_on_done`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - `i32 (bool) completed`
+
+Called when the host is done processing context (`context_id`).
+
+Plugin must return one of the following values:
+- `true` to allow the host to finalize and delete context.
+- `false` to indicate that the context is still being used,
+  and that plugin is going to call [`proxy_done`] later to
+  allow the host to finalize and delete that context.
+
+
+#### `proxy_on_log`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - none
+
+Called after the host is done processing context, but before
+releasing its state.
+
+This can be used e.g. for generating final log entries.
+
+It's called after `true` was returned from [`proxy_on_done`]
+or after a call to [`proxy_done`].
+
+
+#### `proxy_on_delete`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - none
+
+Called when the host removes the context (`context_id`) to signal that
+the plugin should stop tracking it and remove all associated state.
+
+It's called after `true` was returned from [`proxy_on_done`]
+or after a call to [`proxy_done`].
+
+
+### Functions exposed by the host
+
+#### `proxy_done`
+
+* params:
+  - none
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Indicates to the host that the plugin is done processing active
+context.
+
+This should be used after returning `false` in [`proxy_on_done`].
+
+Returned `status` value is:
+- `OK` on success.
+- `NOT_FOUND` when active context was not pending finalization.
+
+
+#### `proxy_set_effective_context`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Changes the effective context to `context_id`.
+
+This can be used to change active context, e.g. during
+[`proxy_on_http_call_response`], [`proxy_on_grpc_receive`]
+and/or [`proxy_on_queue_ready`] callbacks.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `context_id`.
+
+
+## Configuration
+
+### Callbacks exposed by the Wasm module
+
+#### `proxy_on_vm_start`
+
+* params:
+  - `i32 (uint32_t) plugin_context_id`
+  - `i32 (size_t) vm_configuration_size`
+* returns:
+  - `i32 (bool) status`
+
+Called when the host starts the WebAssembly Virtual Machine.
+
+Its configuration (of `vm_configuration_size`) can be retrieved using
+[`proxy_get_buffer_bytes`] with `buffer_id` set to `VM_CONFIGURATION`.
+
+Plugin must return one of the following values:
+- `true` to indicate that the configuration was processed successfully.
+- `false` to indicate that the configuration processing failed.
+
+
+#### `proxy_on_configure`
+
+* params:
+  - `i32 (uint32_t) plugin_context_id`
+  - `i32 (size_t) plugin_configuration_size`
+* returns:
+  - `i32 (bool) status`
+
+Called when the host starts the Proxy-Wasm plugin.
+
+Its configuration (of `plugin_configuration_size`) can be retrieved
+using [`proxy_get_buffer_bytes`] with `buffer_id` set to
+`PLUGIN_CONFIGURATION`.
+
+Plugin must return one of the following values:
+- `true` to indicate that the configuration was processed successfully.
+- `false` to indicate that the configuration processing failed.
+
+
+## Logging
+
+### Functions exposed by the host
+
+#### `proxy_log`
+
+* params:
+  - `i32 (`[`proxy_log_level_t`]`) log_level`
+  - `i32 (const char *) message_data`
+  - `i32 (size_t) message_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Logs message (`message_data`, `message_size`) at the `log_level`.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `log_level`.
+- `INVALID_MEMORY_ACCESS` when `message_data` and/or `message_size`
+  point to invalid memory address.
+
+
+#### `wasi_snapshot_preview1.fd_write`
+
+* params:
+  - `i32 (`[`wasi_fd_id_t`]`) fd_id`
+  - `i32 (wasi_iovec_array_t) iovec`
+  - `i32 (size_t) iovec_size`
+  - `i32 (size_t *) return_written_bytes`
+* returns:
+  - `i32 (`[`wasi_errno_t`]`) errno`
+
+Logs message (`iovec`, `iovec_size`).
+
+When `fd_id` is `STDOUT`, then it's logged at the `INFO` level.
+
+When `fd_id` is `STDERR`, then it's logged at the `ERROR` level.
+
+Returned `errno` value is:
+- `SUCCESS` on success.
+- `BADF` for unknown or unsupported `fd_id`.
+- `FAULT` when `iovec`, `iovec_size` and/or `return_written_bytes`
+  point to invalid memory address.
+
+
+#### `proxy_get_log_level`
+
+* params:
+  - `i32 (`[`proxy_log_level_t`]` *) return_log_level`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves host's current log level (`return_log_level`).
+
+This can be used to avoid creating log entries that are going to be
+discarded by the host.
+
+Returned `status` value is:
+- `OK` on success.
+- `INVALID_MEMORY_ACCESS` when `return_log_level` points to invalid
+  memory address.
+
+
+## Clocks
+
+### Functions exposed by the host
+
+#### `proxy_get_current_time_nanoseconds`
+
+* params:
+  - `i32 (uint64_t *) return_time`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+> **Warning**
+> This function has been deprecated in favor of
+> [`wasi_snapshot_preview1.clock_time_get`].
+
+Retrieves current time (`return_time`), or the approximation of it.
+
+> **Note**
+> Hosts might return approximate time (e.g. frozen at the context
+> creation) to improve performance and/or prevent various attacks,
+> so consumers shouldn't assume that time changes during callbacks.
+
+Returned `status` value is:
+- `OK` on success.
+- `INVALID_MEMORY_ACCESS` when `return_time` points to invalid memory
+  address.
+
+
+#### `wasi_snapshot_preview1.clock_time_get`
+
+* params:
+  - `i32 (`[`wasi_clock_id_t`]`) clock_id`
+  - `i64 (uint64_t) ignored`
+  - `i32 (uint64_t *) return_time`
+* returns:
+  - `i32 (`[`wasi_errno_t`]`) errno`
+
+Retrieves current time (`return_time`), or the approximation of it.
+
+> **Note**
+> Hosts might return approximate time (e.g. frozen at the context
+> creation) to improve performance and/or prevent various attacks,
+> so consumers shouldn't assume that time changes during callbacks.
+
+Returned `errno` value is:
+- `SUCCESS` on success.
+- `NOTSUP` for unknown or unsupported `clock_id`.
+- `FAULT` when `return_time` points to invalid memory address.
+
+
+## Timers
+
+### Functions exposed by the host
+
+#### `proxy_set_tick_period_milliseconds`
+
+* params:
+  - `i32 (uint32_t) tick_period`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets timer period (`tick_period`). When set, the host will call
+[`proxy_on_tick`] every `tick_period` milliseconds.
+
+Returned `status` value is:
+- `OK` on success.
+
+
+### Callbacks exposed by the Wasm module
+
+#### `proxy_on_tick`
+
+* params:
+  - `i32 (uint32_t) plugin_context_id`
+* returns:
+  - none
+
+Called on a timer every tick period.
+
+The tick period can be configured using
+[`proxy_set_tick_period_milliseconds`].
+
+
+## Randomness
+
+### Functions exposed by the host
+
+#### `wasi_snapshot_preview1.random_get`
+
+* params:
+  - `i32 (uint8_t *) buffer`
+  - `i32 (size_t) buffer_size`
+* returns:
+  - `i32 (`[`wasi_errno_t`]`) errno`
+
+Generates `buffer_size` bytes of randomness.
+
+Returned `errno` value is:
+- `SUCCESS` on success.
+- `INVAL` when requested `buffer_size` is too large.
+- `FAULT` when `buffer` and/or `buffer_size` point to invalid memory
+  address.
+
+
+## Environment variables
+
+> **Warning**
+> Environment variables should be configured for each plugin.
+> Exposing the host's environment variables is highly discouraged.
+
+
+### Functions exposed by the host
+
+#### `wasi_snapshot_preview1.environ_sizes_get`
+
+* params:
+  - `i32 (size_t *) return_num_elements`
+  - `i32 (size_t *) return_buffer_size`
+* returns:
+  - `i32 (`[`wasi_errno_t`]`) errno`
+
+Returns `return_num_elements` of environment variables, and
+`return_buffer_size` sufficient to serialize them and decorators.
+
+Returned `errno` value is:
+- `SUCCESS` on success.
+- `FAULT` when `return_num_elements` and/or `return_buffer_size`
+  point to invalid memory address.
+
+
+#### `wasi_snapshot_preview1.environ_get`
+
+* params:
+  - `i32 (uint8_t **) return_array`
+  - `i32 (uint8_t *) return_buffer`
+* returns:
+  - `i32 (`[`wasi_errno_t`]`) errno`
+
+Returns serialized environment variables.
+
+Returned `errno` value is:
+- `SUCCESS` on success.
+- `FAULT` when `return_array` and/or `return_buffer` point to
+  invalid memory address.
+
+
+## Buffers
+
+Access to buffers using functions in this section is restricted to
+specific callbacks:
+
+- `HTTP_REQUEST_BODY` can be read and modified in
+  [`proxy_on_request_body`] (or for as long as request processing
+  is paused from it).
+- `HTTP_RESPONSE_BODY` can be read and modified in
+  [`proxy_on_response_body`] (or for as long as response processing
+  is paused from it).
+- `DOWNSTREAM_DATA` can be read and modified in
+  [`proxy_on_downstream_data`] (or for as long as upstream processing
+  is paused from it).
+- `UPSTREAM_DATA` can be read and modified in
+  [`proxy_on_downstream_data`] (or for as long as downstream processing
+  is paused from it).
+- `HTTP_CALL_RESPONSE_BODY` can be read in
+  [`proxy_on_http_call_response`].
+- `GRPC_CALL_MESSAGE` can be read in [`proxy_on_grpc_receive`].
+- `VM_CONFIGURATION` can be read in [`proxy_on_vm_start`].
+- `PLUGIN_CONFIGURATION` can be read in [`proxy_on_configure`].
+- `FOREIGN_FUNCTION_ARGUMENTS` can be read in
+  [`proxy_on_foreign_function`].
+
+
+### Functions exposed by the host
+
+#### `proxy_set_buffer_bytes`
+
+* params:
+  - `i32 (`[`proxy_buffer_type_t`]`) buffer_id`
+  - `i32 (uint32_t) offset`
+  - `i32 (uint32_t) size`
+  - `i32 (const uint8_t *) value_data`
+  - `i32 (size_t) value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets content of the buffer `buffer_id` to the provided value
+(`value_data`, `value_size`) replacing `size` bytes starting
+at `offset` in the existing buffer.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `buffer_id`.
+- `NOT_FOUND` when requested `buffer_id` isn't available.
+- `INVALID_MEMORY_ACCESS` when `value_data` and/or `value_size`
+  point to invalid memory address.
+
+
+#### `proxy_get_buffer_bytes`
+
+* params:
+  - `i32 (`[`proxy_buffer_type_t`]`) buffer_id`
+  - `i32 (uint32_t) offset`
+  - `i32 (uint32_t) max_size`
+  - `i32 (uint8_t **) return_value_data`
+  - `i32 (size_t *) return_value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves up to` max_size` bytes starting at `offset` from the buffer
+`buffer_id`.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `buffer_id`, or in case of buffer overflow
+   due to invalid `offset` and/or `max_size` values.
+- `NOT_FOUND` when requested `buffer_id` isn't available.
+- `INVALID_MEMORY_ACCESS` when `returned_value_data` and/or
+  `returned_value_size` point to invalid memory address.
+
+
+#### `proxy_get_buffer_status`
+
+* params:
+  - `i32 (`[`proxy_buffer_type_t`]`) buffer_id`
+  - `i32 (size_t *) return_buffer_size`
+  - `i32 (uint32_t *) return_ignored`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves size (`return_buffer_size`) of the buffer `buffer_id`.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `buffer_id`.
+- `NOT_FOUND` when requested `buffer_id` isn't available.
+- `INVALID_MEMORY_ACCESS` when `return_buffer_size` and/or
+  `return_ignored` point to invalid memory address.
+
+
+## HTTP fields
+
+Access to HTTP fields using functions in this section is restricted to
+specific callbacks:
+
+- `HTTP_REQUEST_HEADERS` can be read in [`proxy_on_log`],
+  and read and modified from [`proxy_on_request_headers`]
+  (or for as long as request processing is paused from it).
+- `HTTP_REQUEST_TRAILERS` can be read in [`proxy_on_log`],
+  and read and modified in [`proxy_on_request_trailers`]
+  (or for as long as request processing is paused from it).
+  They can be added in [`proxy_on_request_body`], but only when the
+  proxied request doesn't have trailers (`end_of_stream` is `true`).
+- `HTTP_RESPONSE_HEADERS` can be read in [`proxy_on_log`],
+  and read and modified in [`proxy_on_response_headers`]
+  (or for as long as response processing is paused from it).
+- `HTTP_RESPONSE_TRAILERS` can be read in [`proxy_on_log`],
+  and read and modified in [`proxy_on_response_trailers`]
+  (or for as long as response processing is paused from it).
+  They can be added in [`proxy_on_response_body`], but only when the
+  proxied response doesn't have trailers (`end_of_stream` is `true`).
+- `GRPC_CALL_INITIAL_METADATA` can be read in
+  [`proxy_on_grpc_receive_initial_metadata`].
+- `GRPC_CALL_TRAILING_METADATA` can be read in
+  [`proxy_on_grpc_receive_trailing_metadata`].
+- `HTTP_CALL_RESPONSE_HEADERS` can be read in
+  [`proxy_on_http_call_response`].
+- `HTTP_CALL_RESPONSE_TRAILERS` can be read in
+  [`proxy_on_http_call_response`].
+
+
+### Functions exposed by the host
+
+#### `proxy_get_header_map_size`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (size_t *) return_pairs_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves size (`return_pairs_size`) of all key-value pairs from
+the header map `map_id`.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
+- `INVALID_MEMORY_ACCESS` when `return_pairs_size` points to
+  invalid memory address.
+
+
+#### `proxy_get_header_map_pairs`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (uint8_t **) return_serialized_pairs_data`
+  - `i32 (size_t *) return_serialized_pairs_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves all key-value pairs from the header map `map_id`.
+
+Returned header map (`return_serialized_pairs_data`,
+`return_serialized_pairs_size`) is [serialized].
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
+- `INVALID_MEMORY_ACCESS` when `return_pairs_data` and/or
+  `return_pairs_size` point to invalid memory address.
+
+
+#### `proxy_set_header_map_pairs`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const uint8_t *) serialized_pairs_data`
+  - `i32 (size_t) serialized_pairs_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets all key-value pairs in the header map `map_id` to the provided
+[serialized] map (`serialized_pairs_data`, `serialized_pairs_size`).
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
+- `INVALID_MEMORY_ACCESS` when `pairs_data` and/or
+  `pairs_size` point to invalid memory address.
+
+
+#### `proxy_get_header_map_value`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const char *) key_data`
+  - `i32 (size_t) key_size`
+  - `i32 (uint8_t **) return_value_data`
+  - `i32 (size_t *) return_value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves value (`return_value_data`, `return_value_size`) of the header
+key (`key_data`, `key_value`) from the header map `map_id`.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
+- `NOT_FOUND` when the requested key was not found.
+- `INVALID_MEMORY_ACCESS` when `key_data`, `key_size`,
+  `return_value_data` and/or `return_value_size` point to
+  invalid memory address.
+
+
+#### `proxy_add_header_map_value`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const char *) key_data`
+  - `i32 (size_t) key_size`
+  - `i32 (const uint8_t *) value_data`
+  - `i32 (size_t) value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Adds key (`key_data`, `key_size`) with value (`value_data`,
+`value_size`) to the header map `map_id`.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
+- `INVALID_MEMORY_ACCESS` when `key_data`, `key_size`, `value_data`
+  and/or `value_size` point to invalid memory address.
+
+
+#### `proxy_replace_header_map_value`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const char *) key_data`
+  - `i32 (size_t) key_size`
+  - `i32 (const uint8_t *) value_data`
+  - `i32 (size_t) value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Adds or replaces key's (`key_data`, `key_value`) value with the provided
+value (`value_data`, `value_size`) in the header map `map_id`.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
+- `INVALID_MEMORY_ACCESS` when `key_data`, `key_size`, `value_data`
+  and/or `value_size` point to invalid memory address.
+
+
+#### `proxy_remove_header_map_value`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const char *) key_data`
+  - `i32 (size_t) key_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Removes the key (`key_data`, `key_value`) from the header map `map_id`.
+
+Returned `status` value is:
+- `OK` on success (including the case when requested key didn't exist).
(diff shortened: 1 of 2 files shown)


PiotrSikora:
Please take a look, I'm sure there are some errors, but it should be fairly complete.

I'll wait a week or two before merging this to make sure all mistakes are caught and fixed. Afterwards, I'll push v0.1.0 and v0.2.0 for completeness, since it's going to be trivial to do those.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+* returns:
+  - none
+
+> **Warning**
+> This is called only if [`_initialize`] is not exported.
+
+Called when the Wasm module is first loaded.
+
+
+### `proxy_on_memory_allocate`
```
codefromthecrypt, line 81: maybe somewhere say that either `proxy_on_memory_allocate` or `malloc` may be absent (not exported by the guest), but not both?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+
+### Functions exposed by the host environment
+
+All functions exposed by the host return [`proxy_status_t`], which
+indicates status of the call (success, invalid memory access, etc.),
+and return values are written into memory pointed by `return_<value>`
+parameters.
+
+
+## Integration
```
codefromthecrypt, line 29: maybe "guest exports"? This will make more clear which side is supposed to export these? Or maybe say so as a summary comment below.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+the in-VM memory allocator.
+
+Plugin must return `memory_data` pointing to the start of the allocated
+memory.
+
+Returning `0` indicates failure.
+
+
+## Context lifecycle
+
```
codefromthecrypt, line 117: needs summary comment that everything in this section are exported by the host, and if the host is expected to export all of them.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+[`proxy_on_http_call_response`], [`proxy_on_grpc_receive`]
+and/or [`proxy_on_queue_ready`] callbacks.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `context_id`.
+
+
+## Configuration
+
```
codefromthecrypt, line 210: add summary comment that these are exported by the guest, and if any are required to be exported.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+using [`proxy_get_buffer_bytes`] with `buffer_id` set to
+`PLUGIN_CONFIGURATION`.
+
+Plugin must return one of the following values:
+- `true` to indicate that the configuration was processed successfully.
+- `false` to indicate that the configuration processing failed.
+
+
+## Logging
+
```
codefromthecrypt, line 249: summary comment that this is a host export and if it is required

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `log_level`.
+- `INVALID_MEMORY_ACCESS` when `message_data` and/or `message_size`
+  point to invalid memory address.
+
+
+### `wasi_snapshot_preview1.fd_write`
+
```
codefromthecrypt, line 269: same.. summary comment if this is required, which would be odd. in fact, _start and this may be better put into an "optional support" section unless they are required.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+Returned `errno` value is:
+- `SUCCESS` on success.
+- `BADF` for unknown or unsupported `fd_id`.
+- `FAULT` when `iovec`, `iovec_size` and/or `return_written_bytes`
+  point to invalid memory address.
+
+
+### `proxy_get_log_level`
+
```
codefromthecrypt, line 294: comment this is a guest function, better to but with the host in the same section and just clarify each?

codefromthecrypt (commented):
I started making some comments, then realized they would be the same through the whole doc. Hopefully this gives a good idea of what may be helpful.

abi-versions/v0.2.1/README.md
codefromthecrypt: in general, it would be better to keep all things like this (things that are not defined in this ABI) as a historical appendix. That envoy's host exports a few functions is interesting info, so is that some compilers only have a wasi target. However, even though these topics affect portability, these  are different than the ABI in general which is used by hosts besides envoy such as mosn cc @antjack

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+### `proxy_get_status`
+
+* params:
+  - `i32 (uint32_t *) return_status_code`
+  - `i32 (const char **) return_status_message_data`
+  - `i32 (size_t *) return_status_message_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves gRPC `return_status_code` and `return_status_message`
```
mpwarres, line 1180: Is the `proxy_get_status` hostcall specific to gRPC call statuses? The [GeneralInterface::getStatus](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/c932df12c9db59d331dcb1020174d5f911f67244/include/proxy-wasm/context_interface.h#L566-L570) method in proxy-wasm-cpp-host that corresponds to this has doc that states:
> Provides the status of the last call into the VM or out of the VM, similar to errno.

mpwarres (commented):
This looks great, thanks for writing it!

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Returned `status` value is:
+- `OK` on success.
+- `NOT_FOUND` when there was no property found at the requested `path`.
+- `SERIALIZATION_FAILURE` TODO
+- `INVALID_MEMORY_ACCESS` when `path_data`, `path_size`,
+  `return_value_data` and/or `return_value_size` point to invalid
+  memory address.
```
kyessenov, line 1496: This is incomplete without a description of the encoding of the input / output I think.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+
+- `REALTIME` = `0`
+- `MONOTONIC` = `1`
+
+
+# Serialization
+
+## HTTP fields and gRPC metadata
+
+TODO
```
kyessenov, line 1710: Data encoding is almost as important as the function declarations.

abi-versions/v0.2.1/README.md
PiotrSikora: The initial vNEXT spec had the whole document split into 2 sections (guest and host exports) (e.g. [link](https://github.com/proxy-wasm/spec/tree/442687bb307ea7c701d5744b3f0f3e8308010238/abi-versions/vNEXT#functions-implemented-in-the-wasm-module)).

The improved vNEXT that was under review in #1 had each section split into 2 for guest and host exports (e.g. [link](https://github.com/PiotrSikora/spec/blob/cc53262df056036427476b272fb8f2438aa7975f/abi-versions/vNEXT/README.md#functions-exposed-by-the-host-environment-1)).

I removed those this time for a silly reason that some Markdown tools don't render 4th level headings in the sidebar index, so you couldn't jump right to a specific function, and I assumed that it's pretty easy to tell which is which from the `proxy_on_<event>` function names and descriptions (`Called to/when...`), but I'm happy to re-add those (I'll go with split inside each section, like in the improved vNEXT). Thanks!

abi-versions/v0.2.1/README.md
PiotrSikora: I've added note that all hostcalls are required.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+---
+
+# Callbacks and Functions
+
+### Callbacks exposed by the Wasm module
+
+All callbacks (entry points named `proxy_on_<event>`) are optional,
+and are going to be called only if exposed by the Wasm module.
+
+All callbacks, other than the integration functions, include a unique
```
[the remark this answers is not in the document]
PiotrSikora: Done.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+context identifier as the first parameter, which can be used to
+distinguish between different contexts.
+
+
+### Functions exposed by the host environment
+
+All functions exposed by the host return [`proxy_status_t`], which
```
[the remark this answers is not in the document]
PiotrSikora: Done.

abi-versions/v0.2.1/README.md
PiotrSikora: But those are neither historical not optional, and intercepting `fd_write` is actually quite useful when integrating with existing 3rd party libraries that print to `stdout` and/or `stderr`.

Ideally, all Proxy-Wasm plugins should use proper logging facilitates and use `proxy_log`, but that's not always possible.

abi-versions/v0.2.1/README.md
PiotrSikora: This is a little tricky. One of those callbacks is required in the current Envoy implementation, but technically plugins don't need to export this if they don't retrieve strings from the host.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `context_id`.
+
+
+## Configuration
+
+### `proxy_on_vm_start`
+
+* params:
+  - `i32 (uint32_t) context_id`
```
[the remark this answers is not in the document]
PiotrSikora: I don't believe it is (but I might be mistaken).

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+* params:
+  - `i32 (`[`proxy_buffer_type_t`]`) buffer_id`
+  - `i32 (uint32_t) offset`
+  - `i32 (uint32_t) size`
+  - `i32 (const uint8_t *) value_data`
+  - `i32 (size_t) value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets content of the buffer `buffer_id` to the `value`, replacing
```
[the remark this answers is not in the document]
PiotrSikora: I tried to be "too smart" and imply that `value` is a vector/string represented by `value_data` and `value_size`, but on the second read, you're right that's not very obvious.

I'll go back to using "value (`value_data`, `value_size`)" which hopefully should be sufficient?

Regarding `size` and `value_size`, the former refers to the existing buffer and the latter refers to the provided value. I think that's pretty clear from the existing sentence (especially after adding "value (`value_data`, `value_size`)")?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+
+* params:
+  - `i32 (uint32_t) stream_id`
+  - `i32 (const uint8_t *) message_data`
+  - `i32 (size_t) message_size`
+  - `i32 (bool) end_stream`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sends gRPC `message` on the gRPC stream `stream_id` previously
```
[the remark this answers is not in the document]
PiotrSikora: Done (here, and hopefully throughout the document, so please comment if you notice this in other places).

abi-versions/v0.2.1/README.md
PiotrSikora: That was the original plan, but the only thing this currently implements is gRPC status (and HTTP call failure, but that's hardcoded to a single `"reset"` value, so I ignored it).

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const char *) key_data`
+  - `i32 (size_t) key_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Removes the `key` from the header map `map_id`.
+
+Returned `status` value is:
```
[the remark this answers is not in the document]
PiotrSikora: It does not, calling `proxy_remove_header_map_value(..., "foo", 3)` will return `OK`, since it reaches the desired state and the key named "foo" doesn't exist after this call returns.

I could be convinced that returning `NOT_FOUND` (at least at the ABI layer, SDKs can do whatever feels more idiomatic) would be better, but that's not what's currently implemented.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets all key-value pairs in the header map `map_id` to `pairs`.
+
+`pairs` must be [serialized].
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
```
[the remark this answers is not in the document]
PiotrSikora: Envoy will always return `OK` here, since it doesn't have a concept of restricted headers.

Technically, host could return `BAD_ARGUMENT` for any reason, since it's an expected return value from this function, and I'd use that to support such use case with `v0.2.1`, but I'd prefer not to document and retrofit those concepts into this version of the ABI.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+  - `i32 (const uint8_t *) value_data`
+  - `i32 (size_t) value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Adds `key` with `value` to the header map `map_id`.
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
```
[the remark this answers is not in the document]
PiotrSikora: Same answer ;)

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+- `INVALID_MEMORY_ACCESS` when `returned_value_data` and/or
+  `returned_value_size` point to invalid memory address.
+
+
+### `proxy_get_buffer_status`
+
+* params:
+  - `i32 (`[`proxy_buffer_type_t`]`) buffer_id`
+  - `i32 (size_t *) return_buffer_size`
+  - `i32 (uint32_t *) return_buffer_flags`
```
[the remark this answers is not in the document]
PiotrSikora: Good catch! I replaced this with `return_ignored`.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+
+
+### `proxy_on_new_connection`
+
+* params:
+  - `i32 (uint32_t) context_id`
+* returns:
+  - `i32 (`[`proxy_action_t`]`) action`
+
+Called on a new connection.
```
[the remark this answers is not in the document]
PiotrSikora: Done.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+* returns:
+  - `i32 (`[`proxy_action_t`]`) action`
+
+Called for each data chunk received from upstream.
+
+Data (of `data_size`) can be retrieved and/or modified using
+[`proxy_get_buffer_bytes`] and/or [`proxy_set_buffer_bytes`]
+with `buffer_id` set to `UPSTREAM_DATA`.
+
+Paused downstream can be resumed using [`proxy_continue_stream`]
```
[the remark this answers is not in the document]
PiotrSikora: Done.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+- `INVALID_MEMORY_ACCESS` when `return_buffer_size` and/or
+  `return_buffer_flags` point to invalid memory address.
+
+
+## HTTP fields
+
+### `proxy_get_header_map_size`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
```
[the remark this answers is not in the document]
PiotrSikora: Added note at the beginning of `HTTP fields` section.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (uint32_t) call_id`
+  - `i32 (size_t) num_headers`
+  - `i32 (size_t) body_size`
+  - `i32 (size_t) num_trailers`
+* returns:
+  - none
+
+Called when HTTP response for `call_id` is received.
```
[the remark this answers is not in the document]
PiotrSikora: Done.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+  - `i32 (const uint8_t *) body_data`
+  - `i32 (size_t) body_size`
+  - `i32 (const uint8_t *) serialized_headers_data`
+  - `i32 (size_t) serialized_headers_size`
+  - `i32 (uint32_t) grpc_status`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sends HTTP response with `body` and [serialized] `headers`
+without forwarding the original request upstream.
```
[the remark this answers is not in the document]
PiotrSikora: Good catch, thanks! Updated.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1969 @@
...
+- `HTTP_CALL_RESPONSE_BODY` = `4`
+- `GRPC_CALL_MESSAGE` = `5`
+- `VM_CONFIGURATION` = `6`
+- `PLUGIN_CONFIGURATION` = `7`
+- `FOREIGN_FUNCTION_ARGUMENTS` = `8`
+
+
+#### `proxy_header_map_type_t`
+
+- `HTTP_REQUEST_HEADERS` = `0`
```
thibaultcha, line 1814: Hey @PiotrSikora! Quick question: has it been considered to make URI arguments a map type as well? How are proxy-wasm users currently expected to manipulate that part of the request? Through platform.implementation-specific properties maybe? Would it make sense to introduce a  `HTTP_REQUEST_URI_ARGS` type?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1969 @@
...
+using [`proxy_get_buffer_bytes`] and/or [`proxy_set_buffer_bytes`]
+with `buffer_id` set to `HTTP_RESPONSE_BODY`.
+
+Paused HTTP response can be resumed using [`proxy_continue_stream`]
+or closed using [`proxy_close_stream`] with `stream_type` set to
+`HTTP_RESPONSE`.
+
+Plugin must return one of the following values:
+- `CONTINUE` to forward `HTTP_RESPONSE_BODY` buffer downstream.
+- `PAUSE` to pause processing.
```
thibaultcha, line 1021: In the Rust SDK examples we recently noticed that returning `PAUSE` expects the host implementation to buffer the response body until `eof`, at which point it will call `on_response_body` once more (with a full body and `eof=true`). We saw this described [in this example](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/examples/http_body/src/lib.rs#L53-L55) but not written explicitly so anywhere else. Would this be the right place to clarify this behavior?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1807 @@
...
+
+Plugin must return one of the following values:
+- `true` to indicate that the configuration was processed successfully.
+- `false` to indicate that the configuration processing failed.
+
+
+### `proxy_on_configure`
+
+* params:
+  - `i32 (uint32_t) context_id`
```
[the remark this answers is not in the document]
mpwarres: Thinking about this more, maybe that would be overly verbose. But at the same time it could be helpful to designate which callbacks carry context_ids for root vs. non-root contexts.

I'm thinking in particular of callbacks like `proxy_on_http_call_response`, `proxy_on_grpc_receive`, etc., where IIRC the context_id is for the root context rather than the stream context that initiated the call. It might be good to indicate this somehow, either by using the name `root_context_id` vs. `context_id` to indicate which is passed, or by adding some text about callbacks that are delivered to the root context rather than stream context. WDYT?

mpwarres (approved):
LGTM mod comments, acknowledging some TODOs could be addressed in later PRs.

abi-versions/v0.2.1/README.md
PiotrSikora: Honestly, I'm treating `proxy_{get,set}_property` in v0.2.1 as a opaque pass-thru to CEL (hence the comment at the beginning of this section), and I was hoping not to document existing paths, since the set isn't stable across  
host implementations.

Maybe we could redirect users to the documentation in proxies for CEL attributes?

In any case, I think CEL makes sense for a lot of use-cases (e.g. when used in expressions, or to access unspecified objects like Envoy's filter state or dynamic metadata), but I don't think it works great as a generic property getter for predefined values.

Notably:
1. Querying using strings vs enums is wasteful, and inconsistent with rest of the Proxy-Wasm ABI.
2. The set isn't stable, so it's impossible to enforce it at the ABI layer, and we had issues with that in the past.
3. There is a some overlap with existing functions (e.g. access to header maps), which is source of some confusion.

abi-versions/v0.2.1/README.md
PiotrSikora: Right, I'll add this shortly. I left `TODO` because I didn't want to delay this PR any longer, but I intend to complete this sections before merging this PR.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+[`proxy_replace_header_map_value`]: #proxy_replace_header_map_value
+[`proxy_remove_header_map_value`]: #proxy_remove_header_map_value
+[`proxy_continue_stream`]: #proxy_continue_stream
+[`proxy_close_stream`]: #proxy_close_stream
+[`proxy_on_new_connection`]: #proxy_on_new_connection
+[`proxy_on_downstream_data`]: #proxy_on_downstream_data
+[`proxy_on_downstream_connection_close`]: #proxy_on_downstream_connection_close
+[`proxy_on_upstream_data`]: #proxy_on_upstream_data
+[`proxy_on_upstream_connection_close`]: #proxy_on_upstream_connection_close
+[`proxy_on_request_headers`]: #proxy_on_request_headers
```
mpwarres, line 1941: Should proxy_on_request_metadata and proxy_on_response_metadata also be documented?

PiotrSikora: I've added them while drafting this document, but then I realized that we don't have any access to metadata elements, so those callbacks are pretty much useless, and I've decided not to document them (since they are optional anyway).

mpwarres: SGTM, thanks!

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+  - `i32 (uint32_t) offset`
+  - `i32 (uint32_t) size`
+  - `i32 (const uint8_t *) value_data`
+  - `i32 (size_t) value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets content of the buffer `buffer_id` to the provided value
+(`value_data`, `value_size`) replacing `size` bytes starting
+at `offset` in the existing buffer.
```
mathetake, line 517: I think we should document prepend/append/replace behavior based on the offset/size combination as in: https://github.com/envoyproxy/envoy/blob/a0620080acf46534675bc6484c8127ff1c2d0898/source/extensions/common/wasm/context.cc#L129-L149

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+- `INVALID_MEMORY_ACCESS` when `returned_value_data` and/or
+  `returned_value_size` point to invalid memory address.
+
+
+#### `proxy_get_buffer_status`
+
+* params:
+  - `i32 (`[`proxy_buffer_type_t`]`) buffer_id`
+  - `i32 (size_t *) return_buffer_size`
+  - `i32 (uint32_t *) return_ignored`
```
mathetake, line 555: missing doc for this parameter

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+- `BAD_ARGUMENT` for unknown `buffer_id`.
+- `NOT_FOUND` when requested `buffer_id` isn't available.
+- `INVALID_MEMORY_ACCESS` when `return_buffer_size` and/or
+  `return_ignored` point to invalid memory address.
+
+
+## HTTP fields
+
+Access to HTTP fields using functions in this section is restricted to
+specific callbacks:
```
mathetake, line 572: add reference to `proxy_header_map_type_t` to indicate that the following items correspond to one of the enum values?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+#### `proxy_get_header_map_size`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (size_t *) return_pairs_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves size (`return_pairs_size`) of all key-value pairs from
+the header map `map_id`.
```
mathetake, line 611: could you clarify what `header map` mean here? I believe this includes not only HTT_HEADERS but also TRAILERS.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+#### `proxy_get_header_map_pairs`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (uint8_t **) return_serialized_pairs_data`
+  - `i32 (size_t *) return_serialized_pairs_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves all key-value pairs from the header map `map_id`.
```
mathetake, line 629: same comment on `header map`

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets all key-value pairs in the header map `map_id` to the provided
+[serialized] map (`serialized_pairs_data`, `serialized_pairs_size`).
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
+- `INVALID_MEMORY_ACCESS` when `pairs_data` and/or
+  `pairs_size` point to invalid memory address.
```
mathetake, line 657: `pairs_data` and `pairs_size` should be `serialized_pairs_data` and so on?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+Retrieves all key-value pairs from the header map `map_id`.
+
+Returned header map (`return_serialized_pairs_data`,
+`return_serialized_pairs_size`) is [serialized].
+
+Returned `status` value is:
+- `OK` on success.
+- `BAD_ARGUMENT` for unknown `map_id`.
+- `INVALID_MEMORY_ACCESS` when `return_pairs_data` and/or
+  `return_pairs_size` point to invalid memory address.
```
mathetake, line 638:
```suggestion
- `INVALID_MEMORY_ACCESS` when `return_serialized_pairs_data` and/or
  `return_serialized_pairs_size` point to invalid memory address.
```

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const char *) key_data`
+  - `i32 (size_t) key_size`
+  - `i32 (uint8_t **) return_value_data`
+  - `i32 (size_t *) return_value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves value (`return_value_data`, `return_value_size`) of the header
+key (`key_data`, `key_value`) from the header map `map_id`.
```
mathetake, line 672: same comment on `header map`

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const char *) key_data`
+  - `i32 (size_t) key_size`
+  - `i32 (const uint8_t *) value_data`
+  - `i32 (size_t) value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Adds key (`key_data`, `key_size`) with value (`value_data`,
+`value_size`) to the header map `map_id`.
```
mathetake, line 695: same comment on `header map`

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const char *) key_data`
+  - `i32 (size_t) key_size`
+  - `i32 (const uint8_t *) value_data`
+  - `i32 (size_t) value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Adds or replaces key's (`key_data`, `key_value`) value with the provided
+value (`value_data`, `value_size`) in the header map `map_id`.
```
mathetake, line 716: same comment on `header map`

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+#### `proxy_remove_header_map_value`
+
+* params:
+  - `i32 (`[`proxy_header_map_type_t`]`) map_id`
+  - `i32 (const char *) key_data`
+  - `i32 (size_t) key_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Removes the key (`key_data`, `key_value`) from the header map `map_id`.
```
mathetake, line 734: same comment on `header map`

mathetake (commented):
haven't reviewed all, but flushing the first comments. Thank you so much for your work! @PiotrSikora

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+or gRPC stream or call when called from [`proxy_on_grpc_close`].
+
+Returned `status` value is:
+- `OK` on success.
+- `INVALID_MEMORY_ACCESS` when `return_status_code`,
+  `return_status_message_data` and/or `return_status_message_size`
+  point to invalid memory address.
+
+
+## TCP/UDP/QUIC streams
```
mathetake, line 798: I think we should limit this to only `TCP streams` since UDP/QUIC are not supposed in v0.2.1.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+Removes the key (`key_data`, `key_value`) from the header map `map_id`.
+
+Returned `status` value is:
+- `OK` on success (including the case when requested key didn't exist).
+- `BAD_ARGUMENT` for unknown `map_id`.
+- `INVALID_MEMORY_ACCESS` when `key_data` and/or `key_size` point to
+  invalid memory address.
+
+
+## Common stream operations
```
mathetake, line 743: can we clarify what `Common` mean here? This includes not only TCP but also HTTP, right?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+* params:
+  - `i32 (uint32_t) stream_context_id`
+  - `i32 (`[`proxy_peer_type_t`]`) peer_type`
+* returns:
+  - none
+
+Called when upstream connection is closed.
+
+
+## HTTP
```
mathetake, line 906: For consistency, shouldn't this be `HTTP Steams` (as in `Common stream operations`)

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+
+### Callbacks exposed by the Wasm module
+
+> **Note**
+> The same unique `stream_context_id` is shared by
+> HTTP request and response.
```
mathetake, line 912: trailers as well?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+and the body is buffered by the host and not forwarded downstream.
+
+Paused HTTP response can be resumed using [`proxy_continue_stream`]
+or closed using [`proxy_close_stream`] with `stream_type` set to
+`HTTP_RESPONSE`.
+
+Plugin must return one of the following values:
+- `CONTINUE` to forward `HTTP_RESPONSE_BODY` buffer downstream.
+- `PAUSE` to pause processing.
+
```
mathetake, line 1068: Can we clarify that we cannot use `proxy_send_local_response` at this point since the headers might have already arrived at the downstream?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+with `map_id` set to `HTTP_RESPONSE_TRAILERS`.
+
+Paused request can be resumed using [`proxy_continue_stream`]
+or closed using [`proxy_close_stream`] with `stream_type` set to
+`HTTP_REQUEST`.
+
+Plugin must return one of the following values:
+- `CONTINUE` to forward `HTTP_RESPONSE_TRAILERS` fields downstream.
+- `PAUSE` to pause processing.
+
```
mathetake, line 1096: same comment on `proxy_send_local_response` above.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+  - `i32 (const char *) upstream_name_data`
+  - `i32 (size_t) upstream_name_size`
+  - `i32 (const uint8_t *) serialized_headers_data`
+  - `i32 (size_t) serialized_headers_size`
+  - `i32 (const uint8_t *) body_data`
+  - `i32 (size_t) body_size`
+  - `i32 (const uint8_t *) serialized_trailers_data`
+  - `i32 (size_t) serialized_trailers_size`
+  - `i32 (uint32_t) timeout`
+  - `i32 (uint32_t *) return_call_id`
```
mathetake, line 1144: can we document how `return_call_id` will be used?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+* params:
+  - `i32 (const char *) vm_id_data`
+  - `i32 (size_t) vm_id_size`
+  - `i32 (const char *) name_data`
+  - `i32 (size_t) name_size`
+  - `i32 (uint32_t *) return_queue_id`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Resolves existing shared queue using the provided VM ID (`vm_id_data`,
```
mathetake, line 1505: can we document what `VM ID` mean here?  this is the first appearance.

abi-versions/v0.2.1/README.md
mathetake: > I don't think it works great as a generic property getter for predefined values.

huge +1

mathetake:
cc @spacewander @shukitchan as non-Envoy implementors

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+  - `i32 (uint8_t **) return_value_data`
+  - `i32 (size_t *) return_value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves up to` max_size` bytes starting at `offset` from the buffer
+`buffer_id`.
+
+Returned `status` value is:
+- `OK` on success.
```
spacewander, line 542: Should we doc the behavior when the return buffer is empty? (For example, request without a body)

abi-versions/v0.2.1/README.md
thibaultcha: > We could add a note that calling proxy_on_{request,response}_body with the same buffer size twice in a row implies that the buffer is full, but that's a bit hacky and not currently implemented.

I suppose that is not what the proxy-wasm-rust-sdk example expects, and I don't know how much flexibility there is for v0.2.1 and this corner-case, but what if we made it clear body handlers can always be invoked with more chunks so long as `eof: false`? In short doing something like `vNEXT` `WaitForEndOrFull`, but implied.

1. Handlers return `again` the first time which enables buffering.
2. The same handler expects another invocation with a chunk that may or may not fit in the body buffer.
3. Next body handler invocation:
  3a. If `eof: true`, the full body fits in the buffer on this invocation.
  3b. If `eof: false`, the buffer may or may not be full, but either way the handler can consider this body to only be handled by chunks, and expect more invocations.

In this context, hosts are expected to print appropriate error logs for a full body buffer or a second `again` return value. Filters are them expected to keep track of body handlers invocations themselves on a per-need basis. Or rather, body handlers are expected to always handle chunked bodies as a fail-safe.

Maybe this misses cases I haven't thought of, or maybe it matches existing implementations, I don't know. Plus of course I am unsure how much flexibility there is for existing host and filters of v0.2.1.

By the way, what would happen to the [proxy-wasm-rust-sdk example](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/examples/http_body/src/lib.rs#L52-L56) on Envoy, would the body not fit the buffer? I ought to check but I don't know my way around Envoy very well...

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+- `INVALID_MEMORY_ACCESS` when `status_code_details_data`,
+  `status_code_details_size`, `body_data`, `body_size`, `headers_data`
+  and/or `headers_size` point to invalid memory address.
+
+
+## HTTP calls
+
+### Functions exposed by the host
+
+#### `proxy_http_call`
```
younes-io, line 1132: @PiotrSikora  Is the function `proxy_http_call` the replacement of `proxy_dispatch_http_call` in the [old spec](https://github.com/proxy-wasm/spec/commit/86719f51b29cc1fcdf2d07ae80a6cc7bae0bc99d?short_path=c4e3008#diff-c4e3008315244482bb4516fa4106c38d1028b05c508af70b1bea7a75b4b235e5L653) ?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+
+* params:
+  - `i32 (`[`proxy_metric_type_t`]`) metric_type`
+  - `i32 (const char *) name_data`
+  - `i32 (size_t) name_size`
+  - `i32 (uint32_t *) return_metric_id`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Defines a new metric of type `metric_type` with a name (`name_data`,
```
hishamhm, line 1584: A couple of questions about behaviors that are not specified here:

* What happens if I give it a metric name that was previously defined? Do I get the preexisting id in `return_metric_id`?
   * What happens if I pass a pre-existing name but with a different `metric_type`?
      * Are the names namespaced by type and I get a new metric id?
      * Do I get a `BAD_ARGUMENT` error?
      * Do I get the preexisting id and the type is ignored?
      * Do I get the preexisting id and the type is updated?

* This call creates a metric; how do I dispose of a metric? For other functions such as `proxy_set_property` setting the value to NULL can be assumed to mean "unset", but how to "undefine" and release memory from metrics?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+#### `proxy_set_tick_period_milliseconds`
+
+* params:
+  - `i32 (uint32_t) tick_period`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets timer period (`tick_period`). When set, the host will call
+[`proxy_on_tick`] every `tick_period` milliseconds.
+
```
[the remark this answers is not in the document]
machaval: What happen if the tick_period is 0 is it disable?

abi-versions/v0.2.1/README.md
xr: Shall the actions need to extend a bit and make it more explicit? the Pause will actually convert into `Http::FilterDataStatus::StopIterationAndBuffer` status implicitly which will buffer the response body. more context: https://github.com/tetratelabs/proxy-wasm-go-sdk/issues/414

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+Paused request can be resumed using [`proxy_continue_stream`]
+or closed using [`proxy_close_stream`] with `stream_type` set to
+`HTTP_REQUEST`.
+
+Additionally, instead of forwarding request upstream, a HTTP response
+can be sent using [`proxy_send_local_response`].
+
+Plugin must return one of the following values:
+- `CONTINUE` to forward `HTTP_REQUEST_HEADERS` fields downstream.
+- `PAUSE` to pause processing.
```
xr, line 943: notice a case that the action pause is not actual pausing the processing: https://github.com/tetratelabs/proxy-wasm-go-sdk/issues/425

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+[`proxy_get_header_map_value`], [`proxy_replace_header_map_value`],
+[`proxy_add_header_map_value`] and/or [`proxy_remove_header_map_value`]
+with `map_id` set to `HTTP_REQUEST_HEADERS`.
+
+Paused request can be resumed using [`proxy_continue_stream`]
+or closed using [`proxy_close_stream`] with `stream_type` set to
+`HTTP_REQUEST`.
+
+Additionally, instead of forwarding request upstream, a HTTP response
+can be sent using [`proxy_send_local_response`].
```
xr, line 939: Seems like there are several variations of local responses like:
- sendDirectLocalReply https://github.com/envoyproxy/envoy/blob/50324229d7cf6024c43c7fd4bae1e042beea4728/source/common/http/filter_manager.cc#L1099
- sendLocalReply https://github.com/envoyproxy/envoy/blob/50324229d7cf6024c43c7fd4bae1e042beea4728/source/common/http/filter_manager.cc#L944
- sendLocalReplyViaFilterChain https://github.com/envoyproxy/envoy/blob/50324229d7cf6024c43c7fd4bae1e042beea4728/source/common/http/filter_manager.cc#L1061

Is it possible to make here `proxy_send_local_response` more explicitly to indicate which type specifically it will trigger or specifying clearly in the doc?

htuch:
@PiotrSikora any update on when this might be merged?

PiotrSikora:
> @PiotrSikora any update on when this might be merged?

I'll try to wrap this this week.

jedisct1:
🎉

jedisct1:
@PiotrSikora Is there anything blocking this? Reference documentation for `proxy-wasm` would be super useful for new implementations.

PiotrSikora:
> @PiotrSikora Is there anything blocking this? Reference documentation for `proxy-wasm` would be super useful for new implementations.

Nothing major, I need to address some comments and do a minor cleanup.

It's finally on top of my TODO list, so I should land it in a day or two.

abi-versions/v0.2.1/README.md
PiotrSikora: Same answer.

abi-versions/v0.2.1/README.md
PiotrSikora: It's opaque name, it doesn't mean anything, as long as it's used consistently.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+  - `i32 (bool) status`
+
+Called when the host starts the WebAssembly Virtual Machine.
+
+Its configuration (of `vm_configuration_size`) can be retrieved using
+[`proxy_get_buffer_bytes`] with `buffer_id` set to `VM_CONFIGURATION`.
+
+Plugin must return one of the following values:
+- `true` to indicate that the configuration was processed successfully.
+- `false` to indicate that the configuration processing failed.
```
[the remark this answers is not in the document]
PiotrSikora: Done.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,1996 @@
...
+
+Called when the host starts the Proxy-Wasm plugin.
+
+Its configuration (of `plugin_configuration_size`) can be retrieved
+using [`proxy_get_buffer_bytes`] with `buffer_id` set to
+`PLUGIN_CONFIGURATION`.
+
+Plugin must return one of the following values:
+- `true` to indicate that the configuration was processed successfully.
+- `false` to indicate that the configuration processing failed.
```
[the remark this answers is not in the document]
PiotrSikora: Done.

abi-versions/v0.2.1/README.md
PiotrSikora: > * What happens if I give it a metric name that was previously defined? Do I get the preexisting id in `return_metric_id`?
>   
>   * What happens if I pass a pre-existing name but with a different `metric_type`?
>     
>     * Are the names namespaced by type and I get a new metric id?
>     * Do I get a `BAD_ARGUMENT` error?
>     * Do I get the preexisting id and the type is ignored?
>     * Do I get the preexisting id and the type is updated?

The current behavior (i.e. that matching Envoy) is that there can be different metric types that use the same metric name.

Also, Envoy returns a new `return_metric_id` for each call (even for the same metric name and type), but you should return the same metric ID for the same metric and type combination.

> * This call creates a metric; how do I dispose of a metric? For other functions such as `proxy_set_property` setting the value to NULL can be assumed to mean "unset", but how to "undefine" and release memory from metrics?

Removing of various resources (including metrics) is not well supported in v0.2.1 (which this PR tries to document). In the future, you'd use [proxy_delete_metric](https://github.com/PiotrSikora/spec/blob/review/abi-versions/vNEXT/README.md#proxy_delete_metric).

abi-versions/v0.2.1/README.md
```diff
@@ -1791,6 +1908,66 @@ Returned `errno` value is:
...
+> **Note**
+> The encoding of integers is little-endian.
+
+
+#### Maps with HTTP fields and/or gRPC metadata
+
+Non-empty maps are serialized as:
+- 32-bit integer containing the number of keys in the map,
+- a series of pairs of 32-bit integers containing length of key and value,
+- a series of pairs of bytes containing key and value, separated by `NULL`
```
mpwarres, line 1922: nit: "key and value, separated by `NULL` character" could be misinterpreted as meaning that there is a null char between key and value (only), but not between the pair itself and the next pair.

How about something like, "with each key and value byte sequence terminated by a `NULL` (`0x00`) character"?

mpwarres (approved):
Looks great, thanks for creating this!

abi-versions/v0.2.1/README.md
mpwarres: Might also be good to explicitly state that the terminating NULL characters are not counted in the key/value length values written earlier (even though I think the current language already implies that, strictly speaking).

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,2166 @@
...
+- *Stream Context* - a context in which operations from a specific HTTP
+  request/response pair or TCP stream are executed.
+
+
+# Callbacks and Functions
+
+### Callbacks exposed by the Wasm module
+
+All callbacks (entry points named `proxy_on_<event>`) are optional,
+and are going to be called only if exposed by the Wasm module.
```
kyessenov, line 34: and if the capability is enabled on the host?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,2166 @@
...
+
+- *Host* - a proxy or another environment that executes Proxy-Wasm Plugin
+  inside the WebAssembly Virtual Machine (WasmVM).
+
+- *Hostcall* - a function call from Plugin to Host.
+
+- *Callback* - a function call from Host to Plugin (usually, as a result
+  of an external event).
+
+- *Plugin Context* (also known as *Root Context*) - a context in which
```
kyessenov, line 22: Might help to explain how they relate to the execution context. E.g. whether the same VM is called for many streams or not.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,2166 @@
...
+* params:
+  - `i32 (uint32_t) context_id`
+  - `i32 (uint32_t) parent_context_id`
+* returns:
+  - none
+
+Called when the host creates a new context (`context_id`).
+
+When `parent_context_id` is `0` then a new plugin context is created,
+otherwise a new per-stream context is created.
```
kyessenov, line 160: ... in which case the parent_context_id refers to the plugin context.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,2166 @@
...
+
+* params:
+  - `i32 (`[`proxy_log_level_t`]` *) return_log_level`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Retrieves host's current log level (`return_log_level`).
+
+This can be used to avoid creating log entries that are going to be
+discarded by the host.
```
kyessenov, line 344: note: the log level can be changed dynamically on the host.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,2166 @@
...
+* params:
+  - `i32 (`[`proxy_buffer_type_t`]`) buffer_id`
+  - `i32 (size_t) start`
+  - `i32 (size_t) size`
+  - `i32 (const uint8_t *) value_data`
+  - `i32 (size_t) value_size`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets content of the buffer `buffer_id` to the provided value
```
kyessenov, line 539: I think buffer_id cannot be configuration?

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,2166 @@
...
+### Functions exposed by the host
+
+#### `proxy_continue_stream`
+
+* params:
+  - `i32 (`[`proxy_stream_type_t`]`) stream_type`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Resumes processing of paused `stream_type`.
```
kyessenov, line 783: This has to be called if processing was stopped with StopAndBuffer.

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,2166 @@
...
+
+#### `proxy_on_upstream_connection_close`
+
+* params:
+  - `i32 (uint32_t) stream_context_id`
+  - `i32 (`[`proxy_peer_type_t`]`) peer_type`
+* returns:
+  - none
+
+Called when upstream connection is closed.
```
kyessenov, line 932: closed by either remote or local?

kyessenov (approved):
Thank you - my comments can be addressed in a follow-up.

abi-versions/v0.2.1/README.md
PiotrSikora: Technically, capability restrictions are not the part of the ABI surface (although, we should have error codes for capabilities restricting actions specific parameters, if we ever do that).

Nevertheless, I've added some text.

abi-versions/v0.2.1/README.md
PiotrSikora: Done.

abi-versions/v0.2.1/README.md
PiotrSikora: Added a note. Thanks!

abi-versions/v0.2.1/README.md
```diff
@@ -0,0 +1,2166 @@
...
+### Functions exposed by the host
+
+#### `proxy_set_tick_period_milliseconds`
+
+* params:
+  - `i32 (uint32_t) tick_period`
+* returns:
+  - `i32 (`[`proxy_status_t`]`) status`
+
+Sets timer period (`tick_period`). When set, the host will call
```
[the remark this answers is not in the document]
PiotrSikora: Added.

abi-versions/v0.2.1/README.md
PiotrSikora: Yes, it's going to return `NOT_FOUND` in this case.

abi-versions/v0.2.1/README.md
PiotrSikora: This is already documented in the callbacks that can stop processing.

abi-versions/v0.2.1/README.md
PiotrSikora: Good point. Done. Thanks!

abi-versions/v0.2.1/README.md
PiotrSikora: Yes (hence the `peer_type`). I've added some extra text to describe that value.

abi-versions/v0.2.1/README.md
PiotrSikora: Let's leave the lifecycle documentation for a separate PR or document.

kyessenov (approved):

mpwarres (approved):

PR #73, opened 2024-10-31
send local response: clarify call sequence

kyessenov:
Per discussion in https://github.com/envoyproxy/envoy/pull/36809


abi-versions/v0.2.1/README.md
```diff
@@ -1156,7 +1156,8 @@ Plugin must return one of the following values:
 
 Sends HTTP response with body (`body_data`, `body_size`) and
 [serialized] headers (`serialized_headers_data`,
-`serialized_headers_size`).
+`serialized_headers_size`). The response is sent after the current
+callback completes and triggers response callbacks.
```
PiotrSikora, line 1160: This is the opposite of the intended behavior (which is that `proxy_send_local_response()` is always a final call).

We want plugins to do something like this:
```
proxy_on_request_headers() {
   ...
   return proxy_send_local_response(...);
}
```

and not something like this:
```
proxy_on_request_headers() {
   ...
   proxy_send_local_response(...);
   do_something_very_important();
   return Pause;
}
```


The reason this is deferred in Envoy is because the response callbacks (i.e. `proxy_on_response_headers()` and friends) are called immediately upon sending local response, which leads to re-entrancy problems, but that's another issue (I don't think that plugins should "receive" the local responses they generated themselves).

kyessenov: @mpwarres I think we have to specify it one way or another. How does WaaS behave here?

Issue #71, opened 2024-10-09
Create registries for extension points


PiotrSikora:
We need to provide registries in this repository for various extension points:
- settings (#55),
- properties (#60),
- custom hostcalls and callbacks,
to assign enum-values and avoid duplications across implementations.

Also, SDKs should be able to query Host about them at startup, so that SDKs can refuse to start e.g. if a specific custom hostcall isn't available.

PR #74, opened 2024-12-13
Publish a Proxy-Wasm roadmap

martijneken:
Share in-flight efforts and outline the next high-priority features.

This PR is a request for comment. Please tell us what's missing!


docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+    [proxy-sdk](https://crates.io/crates/proxy-sdk) crate into the Rust SDK.
+*   (Help wanted) Build a Lua SDK using a Lua interpreter.
+    *   Seems quite feasible given projects like
+        [wasm_lua](https://github.com/vvanders/wasm_lua).
+    *   Could replace Envoy's
+        [Lua filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter)
+    *   Could benefit NGINX's Lua-based [OpenResty](https://openresty.org/)
+        ecosystem
+*   (Help wanted) Optimize Rust SDK binary size. It seems compiler dead-code
+    elimination is thwarted by the use of Context traits.
```
martijneken, line 71: Self AI: file issues where missing, especially this one which needs context/details

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+
+<a name="envoy"></a>
+
+## Envoy integration
+
+*   (Q1'25: @mpwarres, @botengyao) Get Envoy's inline wasm filter out of alpha
+    (envoyproxy/envoy#36996). Documentation, security scanning, tests, bug
+    fixes, etc.
+*   (TBD: @mpwarres) Implement the v0.3 Proxy-Wasm ABI.
+*   (Help wanted) Decouple from the thread-local execution model. As wasm
```
johnlanni, line 112: I am very interested in this, as it has a significant meaning. If I understand correctly it allows wasm modules to run an event loop in a separate thread without blocking the Envoy worker threads. This makes it possible to reuse native IO-related libraries from Go/Rust/... SDKs (such as HTTP/Redis/MySQL).

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+    [wasi-keyvalue](https://github.com/WebAssembly/wasi-keyvalue), etc). In the
+    short term, we'd like to define the Proxy-Wasm ABI in
+    [WIT](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md),
+    to understand:
+    *   How do
+        [Proxy-Wasm interfaces](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/main/include/proxy-wasm/context_interface.h)
+        map to components?
+    *   What are the API gaps? How should we evolve Proxy-Wasm to become
+        WASI-compatible? What are good incremental steps?
+    *   Are there any performance gaps?
```
krinkinmu, line 44: @martijneken FYI, it's by no means a ready solution, but I started playing with some kind of shim that converts proxy-wasm plugin into WASI. In the ideal world, in the future once it's more or less ready, we can just use it to build a proxy-wasm plugin code into a wasi-http proxy component. The goal is to smooth migration from proxy-wasm to WASI by providing some level of backward compatibility for existing proxy-wasm code.

If you think it's interesting, you can find work in progress in https://github.com/krinkinmu/wasi-http and I'm happy to hear any feedback you may have on the approach and on the details of implementation as well.

mikemorris:
This whole roadmap is really exciting, appreciate the effort in compiling all these initiatives and excited to make some progress here!

johnlanni:
I want to add a task: enhance the security of doAfterVmCallActions, as the current mechanism may lead to unexpected calls on the Host side, resulting in envoy crash(refer to https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/326).

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+
+*   (Q1'25: @piotrsikora, @mpwarres) Publish ABI v0.3, containing at least:
+    *   Feature negotiation (proxy-wasm/spec#71 and proxy-wasm/spec#56)
+    *   Better header/body buffering support (proxy-wasm/spec#63)
+    *   Async shared data (proxy-wasm/spec#54)
+    *   Repeated header support (proxy-wasm/spec#53)
+*   (Help wanted) WASI convergence. We want to adopt the component model at WASI
+    1.0. There is a lot of overlap between Proxy-Wasm and some WASI proposals
+    ([wasi-http](https://github.com/WebAssembly/wasi-http),
+    [wasi-keyvalue](https://github.com/WebAssembly/wasi-keyvalue), etc). In the
```
PiotrSikora, line 35: The WASI convergence was the original goal, but right now (after component model, pre WASI 1.0), those are distinct sets of ABIs, so it's really a two questions:
1) Can WASI ABIs be used in place of Proxy-Wasm ABIs in inline proxies (think performance and low-level integrations)?
2) What is Proxy-Wasm if the ABIs are replaced with WASI? Right now, it's a set of ABIs and their implementations, but none of those will exist after the replacement.

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+    short term, we'd like to define the Proxy-Wasm ABI in
+    [WIT](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md),
+    to understand:
+    *   How do
+        [Proxy-Wasm interfaces](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/main/include/proxy-wasm/context_interface.h)
+        map to components?
```
PiotrSikora, line 41: The `ContextInterface` is an interface between Proxy-Wasm C++ Host and proxies embedding it, it's not a Proxy-Wasm interface.

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+    *   What are the API gaps? How should we evolve Proxy-Wasm to become
+        WASI-compatible? What are good incremental steps?
+    *   Are there any performance gaps?
+*   (Help wanted) Evaluate uses of foreign functions to identify feature gaps.
+    *   For example, Envoy
+        [registers foreign functions](https://github.com/search?q=repo%3Aenvoyproxy%2Fenvoy%20RegisterForeignFunction&type=code)
+        for signature checking, compression, filter state, route cache, and CEL
+        expressions.
+    *   Are there similar extensions in Nginx? Apache Traffic Server?
+    *   Which of these features should be promoted to ABI interfaces?
```
PiotrSikora, line 51: Most (all?) all of them should be promoted incorporated into Proxy-Wasm ABI, definitely those that make sense across multiple implementations.

The question is what are the trade-offs for keeping them as optional registered FFIs (proxy-wasm/spec#71) vs native ABIs?

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+<a name="sdks"></a>
+
+## SDKs / language support
+
+*   (Q1'25: @leonm1) Fork the abandoned Go SDK + support full Golang.
+*   (Google exploring) Build a Python SDK using a MicroPython port.
+*   (Help wanted) Stop using Emscripten in the C++ SDK. Instead use Clang /
+    wasi-sdk (proxy-wasm/proxy-wasm-cpp-sdk#167).
+*   (Help wanted) Merge LeakSignal's
+    [proxy-sdk](https://crates.io/crates/proxy-sdk) crate into the Rust SDK.
```
PiotrSikora, line 62: Those two are low-level implementation details with existing issues, so I'm not sure if there is any value in putting it here? We probably don't want to turn this file into an issue tracker.

docs/Roadmap.md
PiotrSikora, line 71: Same as above (this should be an issue in Rust SDK, not a roadmap item).

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+    ability to swap between multiple wasm runtimes. Help users make an informed
+    decision by benchmarking cold start and execution costs across runtimes.
+*   (Help wanted) Adopt CPU metering as a first-class feature. Leverage
+    instruction counting where available. For other engines (e.g. v8), use a
+    watchdog thread.
+*   (Help wanted) Support dynamic (per VM) limits for RAM and CPU.
+*   (Help wanted) Expand the use of SharedArrayBuffer to reduce memcpy into wasm
+    runtimes. This is especially promising for HTTP body chunks. See relevant
+    [WASI issue](https://github.com/WebAssembly/WASI/issues/594). Also reduce
+    [binary memcpy](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/21a5b089f136712f74bfa03cde43ae8d82e066b6/src/v8/v8.cc#L272).
```
PiotrSikora, line 96: This is slightly more complicated than it initially appears... Supporting this and eliminating copies would require host (including it's modules/extensions) to read/write data from/to the network to/from memory shared with WasmVM, which is linear memory block that needs to be pre-allocated upfront.

PiotrSikora (commented):
Two high-level comments:
1) While I agree that we need to publish a (high-level) roadmap, a lot of items here (especially for the host and Envoy integration) are effectively implementation issues that should be tracked in the issue tracker and not here.
2) This roadmap completely ignores hosts other than Envoy, so it would be great to rewrite some of those issues into proxy-agnostic way (where applicable, but see above).

docs/Roadmap.md
alexsnaps: Anything in particular, from a highlevel/API perspective, that's driving this "merge", the Rust one? Is that "just" low-level implementation details, which is important too, or is there something else that would change the contract with users here? More impacting to the roadmap then I guess.

Looks like tho, [with the acquisition?](https://www.f5.com/company/blog/strengthening-data-protection-and-governance-for-ai-applications), that [repo is gone](https://github.com/leaksignal/proxy-sdk)... source is obviously still in crates.io (under ASLv2).

PiotrSikora: The authors reached out and wanted to contribute their version of the SDK. We had a meeting or two and they were supposed to upstream their changes to the existing SDK via a series of PRs to get them reviewed and to convert them to a supplemental changes to avoid breaking existing users... but that was in May 2024, and nothing materialized since then, so perhaps it should be removed from this roadmap.

As far as I recall, this was primarily driven by their documentation, since it's missing in the current Rust SDK.

README.md
```diff
@@ -14,7 +14,10 @@ Proxy-Wasm extensions across different proxies.
 
 The latest and widely implemented version of the specification is [v0.2.1].
 
+The envisioned evolution for Proxy-Wasm is described in the [roadmap].
```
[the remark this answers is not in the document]
martijneken: How short term it is depends on how much work is put in. I'd argue there's plenty of 'evolution' discussed here.

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
+# Proxy-Wasm Roadmap
+
+The Proxy-Wasm community and maintainers envision an evolution path that has the
+following tracks:
+
+*   [Spec / ABI](#abi)
+*   [SDKs / language support](#sdks)
+*   [Host features](#host)
+*   [Envoy integration](#envoy)
```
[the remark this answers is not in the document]
martijneken: Done

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+response to community interest and contributions.
+
+The overarching goals of this document are to:
+
+*   Publish areas of current investment.
+*   Encourage external contributors by pointing out feature gaps.
+*   Align this repository with the vision of WebAssembly: a portable technology
+    that is cross-language, cross-platform, and cross-provider.
+
+<a name="abi"></a>
```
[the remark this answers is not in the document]
martijneken: I added these to have short/stable IDs for linking, regardless of section headers.

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Spec / ABI
+
+*   (Q1'25: @piotrsikora, @mpwarres) Publish ABI v0.3, containing at least:
+    *   Feature negotiation (proxy-wasm/spec#71 and proxy-wasm/spec#56)
+    *   Better header/body buffering support (proxy-wasm/spec#63)
+    *   Async shared data (proxy-wasm/spec#54)
```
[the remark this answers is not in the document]
martijneken: ACK, removed.

docs/Roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+    *   Repeated header support (proxy-wasm/spec#53)
```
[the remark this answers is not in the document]
martijneken: Done

docs/Roadmap.md
martijneken: Right. It was the only official 'segmentation' of the ABI I could find, but maybe it's distracting. Removed.

docs/Roadmap.md
martijneken: Updated a bit to speak to these questions.

docs/Roadmap.md
martijneken: Awesome! Added a link here.

docs/Roadmap.md
martijneken: Updated wording.

docs/Roadmap.md
martijneken: Ah darn, no wonder we lost touch with them. Removed this.

docs/Roadmap.md
martijneken: Yeah, removed.

docs/Roadmap.md
martijneken: Huh interesting point. AIUI proxy-wasm already has some support for scheduling work (e.g. event callbacks, timers, shared queues), but I think it all ends up back on the calling VM thread or the Envoy main thread (via singletons).

Our interest was mostly about decoupling wasm CPU (so running wasm doesn't block Envoy threads) and limiting overheads (little-used plugins don't need a VM per Envoy worker thread). So we imagined a similar model as today (streams/requests sticky to independent threads), but a manager component that scales each plugin's threads up/down dynamically. The major risk in all of this is blowing up CPU/NUMA cache locality and incurring CPU scheduling overhead.

IIUC you'd like to take it further and give the wasm runtime access to a shared thread pool? This would require new primitives for scheduling work and/or launching threads?
https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#-threading-built-ins

This would be very interesting to discuss in the next community meeting.

docs/Roadmap.md
martijneken: Yeah, perhaps we shouldn't target zero copies, but reducing copies, especially for mostly-readonly content.

README.md
```diff
@@ -14,7 +14,10 @@ Proxy-Wasm extensions across different proxies.
...
+
 [v0.2.1]: abi-versions/v0.2.1/README.md
+[roadmap]: docs/Roadmap.md
 
 ## Implementations
```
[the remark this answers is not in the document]
martijneken: Done

ROADMAP.md
```diff
@@ -0,0 +1,116 @@
...
+    ability to swap between multiple wasm runtimes. Help users make an informed
+    decision by benchmarking cold start and execution costs across runtimes.
+-   (Help wanted) Adopt CPU metering as a first-class feature. Leverage
+    instruction counting where available. For other engines (e.g. v8), use a
+    watchdog thread.
+-   (Help wanted) Support dynamic (per VM) limits for RAM and CPU.
+-   (Help wanted) Expand the use of SharedArrayBuffer to reduce memcpy into wasm
+    runtimes. This is promising for HTTP body chunks (see relevant
+    [WASI issue](https://github.com/WebAssembly/WASI/issues/594)) and
+    [wasm binaries](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/21a5b089f136712f74bfa03cde43ae8d82e066b6/src/v8/v8.cc#L272).
```
PiotrSikora, line 92: I'm not sure why this is linked here?

ROADMAP.md
```diff
@@ -0,0 +1,116 @@
...
+-   (Help wanted) Performance benchmarks. One of Proxy-Wasm's strengths is its
+    ability to swap between multiple wasm runtimes. Help users make an informed
+    decision by benchmarking cold start and execution costs across runtimes.
+-   (Help wanted) Adopt CPU metering as a first-class feature. Leverage
+    instruction counting where available. For other engines (e.g. v8), use a
+    watchdog thread.
+-   (Help wanted) Support dynamic (per VM) limits for RAM and CPU.
+-   (Help wanted) Expand the use of SharedArrayBuffer to reduce memcpy into wasm
+    runtimes. This is promising for HTTP body chunks (see relevant
+    [WASI issue](https://github.com/WebAssembly/WASI/issues/594)) and
```
PiotrSikora, line 91: This is unlikely to ever work, without host being specifically architected to read data from the network directly into Wasm's linear memory block.

AFAIK, the "multiple memories" proposal doesn't allow to attach / detach memories on-the-fly.

PiotrSikora (commented):
This PR should be moved to the [community](https://github.com/proxy-wasm/community) repo, now that we have it.

ROADMAP.md
```diff
@@ -0,0 +1,116 @@
...
+    fixes, etc.
+-   (@mpwarres) Implement the v0.3 Proxy-Wasm ABI.
+-   (Help wanted) Decouple from the thread-local execution model. As wasm
+    modules become more CPU intensive and leverage multiple async APIs, consider
+    managing a separate Proxy-Wasm threadpool. Each VM needs a work queue, and
+    requests need affinity to a single VM. This architecture allows for
+    independent thread scaling (expensive wasms get more CPU), improved
+    parallelism (multiple requests' wasm at the same time), and reduced memory
+    costs (one VM serves multiple Envoy threads). It adds performance risks (CPU
+    scheduling latency, CPU cache misses, NUMA hopping).
```
oschaaf, line 112: drive-by comment - I would love this as an option, but I'm not sure I would love it as a one-size-fits-all due to the concerns mentioned. Thinking out loud; I wonder if the problem could be addressed by offering the decoupling as an api to modules instead, something along the lines of that modules could opt in to posting work on a threadpool (and register for completions or some such)?

Issue #88, opened 2025-06-25
Feature Request: Access to additional properties of peer certificates


DerekTBrown:
## Usecase
- I am attempting to write a WASM Proxy plugin that accesses [x509 extensions](https://docs.redhat.com/en/documentation/red_hat_certificate_system/9/html/administration_guide/standard_x.509_v3_certificate_extensions) of the client certificate.
- Currently, only a limited set of certificate properties are supported. This does not include x509 extensions: 
    - https://github.com/proxy-wasm/spec/blob/155da15d93ac2c56537711586b925f193734122a/abi-versions/v0.2.1/README.md?plain=1#L1769-L1781

## Proposal
I could forsee two paths to implement this behavior. I wanted the community's input on which approach to take:

### **Option A:** Expose x509 Extensions via Property
A narrow solution to this problem would propose the inclusion of x509 extension values via the properties API:
```
 auto buf = getProperty({"connection", "x509_extensions_peer_certificate"});
 if (buf.has_value()) {
      for (const auto& [key, val] : buf.value()->pairs()) {
          LOG_INFO("Got extension": key+" - "+value);
      }
  }
  ```

**Pros**:
- Higher performance than Option B.

**Cons**
- Implementing passthrough functionality to each attribute of a certificate is likely not maintainable in the long-term. I could easily imagine many other similar requests in the future for different attributes of the certificate.

### **Option B:** Expose DER-encoded certificate via properties
A broader solution to this problem would provide a direct DER interface to the underlying certificate via the `getProperty` interface:

```
 auto buf = getProperty({"connection", "der_peer_certificate"});
 if (buf.has_value()) {
      // Parse the DER-encoded certificate on the plugin side.
  }
```

**Pros**
- This is much more maintainable than Option A, since users can parse whatever properties they need out of the certificate.

**Cons**
- This is less performant than Option A:
   - We have to pay a CPU cost to serialize the string into DER format.
   - We have to copy the entire DER as a string, which is expensive.
   - We have to pay additional DER cost to deserialize the DER format on the WASM plugin side, when the host has presumably already deserialized the certificate.

## Closing Notes
- Thanks in advance for all your feedback!
- I am new to the proxy-wasm project, so let me know if there are other options that should be considered.

kyessenov:
I think it option B would probably work for more hosts, but on Envoy, you might already have it parsed, and option A would be more efficient. Is this for OIDs? It probably depends if it's a standard extension or not.

[...]

PR #89, opened 2025-07-09
Add certificate OID map properties to vNEXT spec

DerekTBrown:
This PR addresses issue #88 by adding new properties to access certificate OIDs:

  - `connection.oid_map_local_certificate`
  - `connection.oid_map_peer_certificate`
  - `upstream.oid_map_local_certificate`
  - `upstream.oid_map_peer_certificate`

  These properties return maps of all OIDs to their corresponding values in certificates,
  allowing plugins to access all certificate properties including X.509 extensions that may not
  be exposed through the existing certificate properties.

  This enhancement will enable Proxy-WASM plugins to access additional certificate attributes
  needed for advanced use cases, particularly when working with certificates that contain
  non-standard X.509 extensions with custom OIDs.

### DIFF
--- a/abi-versions/vNEXT/README.md
+++ b/abi-versions/vNEXT/README.md
@@ -1780,6 +1780,10 @@ Returned `status` value is:
   the peer certificate
 * `connection.sha256_peer_certificate_digest` (string) - SHA256 digest of
   the peer certificate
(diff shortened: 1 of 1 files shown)


DerekTBrown:
I wrote a prototype implementation for Envoy, demonstrating feasibility:

https://github.com/envoyproxy/envoy/pull/40178

PR #91, opened 2025-09-25
Add support for host features.

### LINKED ISSUE #55: Allow hosts to advertise features/settings at runtime
The ABI version currently enforces only the function signatures, but we have no way to change the meaning of various parameters, or negotiate settings at runtime.

### LINKED ISSUE #56: Make most/all hostcalls optional
Currently, the ABI is enforced at linking time, which means that all hostcalls must be exported by the host. Theoretically, compilers should optimize away the unused hostcalls, but that hasn't be the case, which broke our ability to add new features without releasing new ABI versions.

PiotrSikora:
The `proxy_get_host_features` hostcall can be used by the SDKs to
determine which features are supported by the host environment.

### DIFF
--- a/abi-versions/vNEXT/HOST_FEATURES.md
+++ b/abi-versions/vNEXT/HOST_FEATURES.md
@@ -1,23 +1,24 @@
 # Host features
 
-| Identifier | Name                     | Values                                                                                             | Deprecated |
-|:----------:|:-------------------------|:---------------------------------------------------------------------------------------------------|:----------:|
-|   0x0001   | `HAS_CORE`               | When present, all baseline functions considered essential are supported.                           |     No     |
-|   0x0002   | `HAS_LOGGING`            | When present, all baseline functions for logging are supported.                                    |     No     |
-|   0x0003   | `HAS_HTTP_HEADERS`       | When present, only baseline functions related to HTTP headers are supported.                       |     No     |
-|   0x0004   | `HAS_HTTP_WITH_BODY`     | When present, all baseline functions for HTTP are supported.                                       |     No     |
-|   0x0005   | `HAS_HTTP_CALLS`         | When present, all baseline functions for HTTP calls are supported.                                 |     No     |
-|   0x0006   | `HAS_GRPC_CALLS`         | When present, all baseline functions for gRPC calls are supported.                                 |     No     |
-|   0x0007   | `HAS_GRPC_STREAMS`       | When present, all baseline functions for gRPC streams are supported.                               |     No     |
-|   0x0008   | `HAS_TCP_FILTER`         | When present, only baseline functions related to new TCP connections are supported.                |     No     |
-|   0x0009   | `HAS_TCP_WITH_PAYLOAD`   | When present, all baseline functions for TCP are supported.                                        |     No     |
-|   0x000A   | `HAS_KEY_VALUE_STORES`   | When present, all baseline functions for key-value stores are supported.                           |     No     |
-|   0x000B   | `HAS_SHARED_QUEUES`      | When present, all baseline functions for shared queues are supported.                              |     No     |
-|   0x000C   | `HAS_TIMERS`             | When present, all baseline functions for timers are supported.                                     |     No     |
-|   0x000D   | `HAS_METRICS`            | When present, all baseline functions for metrics are supported.                                    |     No     |
-|   0x000E   | `HAS_PROPERTIES`         | When present, all baseline functions for properties are supported.                                 |     No     |
-|   0x000F   | `HAS_CUSTOM_FUNCTIONS`   | When present, all baseline functions for custom functions are supported.                           |     No     |
-|   0x0F01   | `HAS_WASI_PREVIEW1_CORE` | When present, all baseline functions considered essential from WASI Preview1 are supported.        |     No     |
+| Identifier | Name                           | Value                                                                                              | Deprecated |
+|:----------:|:-------------------------------|:---------------------------------------------------------------------------------------------------|:----------:|
+|   0x0001   | `HAS_CORE`                     | When present, all baseline functions considered essential are supported.                           |     No     |
+|   0x0002   | `HAS_LOGGING`                  | When present, all baseline functions for logging are supported.                                    |     No     |
+|   0x0003   | `HAS_HTTP_HEADERS`             | When present, only baseline functions related to HTTP headers are supported.                       |     No     |
+|   0x0004   | `HAS_HTTP_WITH_BODY`           | When present, all baseline functions for HTTP are supported.                                       |     No     |
+|   0x0005   | `HAS_TCP_FILTER`               | When present, only baseline functions related to new TCP connections are supported.                |     No     |
+|   0x0006   | `HAS_TCP_WITH_PAYLOAD`         | When present, all baseline functions for TCP are supported.                                        |     No     |
+|   0x0007   | `HAS_CALLOUTS_HTTP_BUFFERED`   | When present, all baseline functions for buffered HTTP callouts are supported.                     |     No     |
+|   0x0008   | `HAS_CALLOUTS_HTTP_STREAMING`  | When present, all baseline functions for streaming HTTP callouts are supported.                    |     No     |
+|   0x0009   | `HAS_CALLOUTS_GRPC_BUFFERED`   | When present, all baseline functions for buffered gRPC callouts are supported.                     |     No     |
+|   0x000A   | `HAS_CALLOUTS_GRPC_STREAMING`  | When present, all baseline functions for streaming gRPC callouts are supported.                    |     No     |
+|   0x000B   | `HAS_KEY_VALUE_STORES`         | When present, all baseline functions for key-value stores are supported.                           |     No     |
+|   0x000C   | `HAS_SHARED_QUEUES`            | When present, all baseline functions for shared queues are supported.                              |     No     |
+|   0x000D   | `HAS_TIMERS`                   | When present, all baseline functions for timers are supported.                                     |     No     |
+|   0x000E   | `HAS_METRICS`                  | When present, all baseline functions for metrics are supported.                                    |     No     |
+|   0x000F   | `HAS_PROPERTIES`               | When present, all baseline functions for properties are supported.                                 |     No     |
+|   0x0010   | `HAS_CUSTOM_FUNCTIONS`         | When present, all baseline functions for custom functions are supported.                           |     No     |
+|   0x0F01   | `HAS_WASI_PREVIEW1_CORE`       | When present, all baseline functions considered essential from WASI Preview1 are supported.        |     No     |
 
 
 Identifiers below 0x2000 are reserved for standardized features and options.
--- a/abi-versions/vNEXT/HOST_FEATURES.md
+++ b/abi-versions/vNEXT/HOST_FEATURES.md
@@ -2,22 +2,22 @@
 
 | Identifier | Name                     | Values                                                                                             | Deprecated |
 |:----------:|:-------------------------|:---------------------------------------------------------------------------------------------------|:----------:|
-|   0x0001   | `HAS_CORE`               | When `1`, all baseline functions considered essential are supported.                               |     No     |
-|   0x0002   | `HAS_LOGGING`            | When `1`, all baseline functions for logging are supported.                                        |     No     |
-|   0x0003   | `HAS_HTTP_HEADERS`       | When `1`, only baseline functions related to HTTP headers are supported.                           |     No     |
-|   0x0004   | `HAS_HTTP_WITH_BODY`     | When `1`, all baseline functions for HTTP are supported.                                           |     No     |
-|   0x0005   | `HAS_HTTP_CALLS`         | When `1`, all baseline functions for HTTP calls are supported.                                     |     No     |
-|   0x0006   | `HAS_GRPC_CALLS`         | When `1`, all baseline functions for gRPC calls are supported.                                     |     No     |
-|   0x0007   | `HAS_GRPC_STREAMS`       | When `1`, all baseline functions for gRPC streams are supported.                                   |     No     |
-|   0x0008   | `HAS_TCP_FILTER`         | When `1`, only baseline functions related to new TCP connections are supported.                    |     No     |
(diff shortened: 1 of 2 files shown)


PiotrSikora:
As mentioned in the meeting, this is still missing text about hosts generating `UNIMPLEMENTED` stubs for any imports starting with `proxy_` to avoid linking-time issues during loading phase for imports supported/expected by guests but not implemented yet by the host.

abi-versions/vNEXT/HOST_FEATURES.md
```diff
@@ -0,0 +1,23 @@
+# Host features
+
+| Identifier | Name                     | Values                                                                                             | Deprecated |
+|:----------:|:-------------------------|:---------------------------------------------------------------------------------------------------|:----------:|
+|   0x0001   | `HAS_CORE`               | When `1`, all baseline functions considered essential are supported.                               |     No     |
```
mpwarres: I agree it's best to be explicit, but just out of curiosity, is there ever a case where this bit wouldn't be set?

abi-versions/vNEXT/HOST_FEATURES.md
```diff
@@ -0,0 +1,23 @@
...
+|   0x0002   | `HAS_LOGGING`            | When `1`, all baseline functions for logging are supported.                                        |     No     |
+|   0x0003   | `HAS_HTTP_HEADERS`       | When `1`, only baseline functions related to HTTP headers are supported.                           |     No     |
+|   0x0004   | `HAS_HTTP_WITH_BODY`     | When `1`, all baseline functions for HTTP are supported.                                           |     No     |
+|   0x0005   | `HAS_HTTP_CALLS`         | When `1`, all baseline functions for HTTP calls are supported.                                     |     No     |
+|   0x0006   | `HAS_GRPC_CALLS`         | When `1`, all baseline functions for gRPC calls are supported.                                     |     No     |
+|   0x0007   | `HAS_GRPC_STREAMS`       | When `1`, all baseline functions for gRPC streams are supported.                                   |     No     |
```
mpwarres: I'm not sure if it's worth it to separate streaming from unary gRPC, it's hard for me to imagine a host supporting only non-streaming gRPC.

abi-versions/vNEXT/HOST_FEATURES.md
```diff
@@ -0,0 +1,23 @@
...
+|   0x000B   | `HAS_SHARED_QUEUES`      | When `1`, all baseline functions for shared queues are supported.                                  |     No     |
+|   0x000C   | `HAS_TIMERS`             | When `1`, all baseline functions for timers are supported.                                         |     No     |
+|   0x000D   | `HAS_METRICS`            | When `1`, all baseline functions for metrics are supported.                                        |     No     |
+|   0x000E   | `HAS_PROPERTIES`         | When `1`, all baseline functions for properties are supported.                                     |     No     |
+|   0x000F   | `HAS_CUSTOM_FUNCTIONS`   | When `1`, all baseline functions for custom functions are supported.                               |     No     |
+|   0x0F01   | `HAS_WASI_PREVIEW1_CORE` | When `1`, all baseline functions considered essential from WASI Preview1 are supported.            |     No     |
+
+
+Identifiers below 0x2000 are reserved for standardized features and options. Random numbers above that range should be used for private extensions.
```
mpwarres: Do you expect that when a new feature intended for eventual standardization is added, it would start off with an ID above 0x2000 while still in experimental state, and then later switch to a standard ID?

abi-versions/vNEXT/HOST_FEATURES.md
```diff
@@ -0,0 +1,23 @@
...
+|   0x000A   | `HAS_KEY_VALUE_STORES`   | When `1`, all baseline functions for key-value stores are supported.                               |     No     |
+|   0x000B   | `HAS_SHARED_QUEUES`      | When `1`, all baseline functions for shared queues are supported.                                  |     No     |
+|   0x000C   | `HAS_TIMERS`             | When `1`, all baseline functions for timers are supported.                                         |     No     |
+|   0x000D   | `HAS_METRICS`            | When `1`, all baseline functions for metrics are supported.                                        |     No     |
+|   0x000E   | `HAS_PROPERTIES`         | When `1`, all baseline functions for properties are supported.                                     |     No     |
+|   0x000F   | `HAS_CUSTOM_FUNCTIONS`   | When `1`, all baseline functions for custom functions are supported.                               |     No     |
+|   0x0F01   | `HAS_WASI_PREVIEW1_CORE` | When `1`, all baseline functions considered essential from WASI Preview1 are supported.            |     No     |
```
mpwarres: With `HAS_WASI_PREVIEW1_CORE`, is the host agreeing to provide at least stub hostcalls for all WASI Preview1 hostcalls, or is it only exporting a selected subset of the WASIp1 hostcalls?

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: Keeping the value part leaves the doors open for building upon this mechanism, instead of adding another one in the future, and I think we'll use it sooner rather than later.

Some things that come to mind would be:
1) Some features might need to carry the value, e.g. `MAXIMUM_BUFFER_SIZE, 16384`,
2) In case we want to allow "supported, but not enabled" use case, which would allow plugins to enable/disable some features at startup, e.g. `proxy_set_host_features(HAS_CUSTOM_FUNCTIONS, 1)`.

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: I disagree. Take Envoy for example, it currently supports only fully buffered (i.e. unary) HTTP callouts, so if for some reason we decided to rebuild gRPC support directly on top of HTTP callouts in SDKs, instead of relying on hosts to support gRPC, then we'd lose gRPC streams for callouts.

This is a hypothetical situation, and in such case we wouldn't have those features advertised here, but I'm just trying to show that some host features are not as well supported as they should be, and for example we don't have streaming HTTP callout because they're not supported in Envoy.

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: With any of the features here, host is advertising a working implementations.

Stubs are always expected for `proxy_*` and `wasi_preview1_*` to support optional features (see: https://github.com/proxy-wasm/spec/pull/91#issuecomment-3336431862), but the advertised features are expected to be fully implemented.

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: Correct. This is the same mechanism that's used by IETF and it works quite well during development, since backwards compatibility can be ignored when people are still working on implementations, because it's easy to switch to another random ID without any consensus.

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: Another use case that we might want to use is to define `BASELINE_VERSION` with random (or monotonic?) values while we're actively working on v0.3 implementations, before we freeze it as final `BASELINE_VERSION=3000` or similar.

abi-versions/vNEXT/HOST_FEATURES.md
```diff
@@ -0,0 +1,23 @@
+# Host features
+
+| Identifier | Name                     | Values                                                                                             | Deprecated |
+|:----------:|:-------------------------|:---------------------------------------------------------------------------------------------------|:----------:|
+|   0x0001   | `HAS_CORE`               | When `1`, all baseline functions considered essential are supported.                               |     No     |
+|   0x0002   | `HAS_LOGGING`            | When `1`, all baseline functions for logging are supported.                                        |     No     |
```
leonm1: Are the SDKs expected to track which host features are supported and avoid invoking their hostcalls, or is it up to plugin authors to avoid calling such functions?

abi-versions/vNEXT/README.md
```diff
@@ -652,6 +725,10 @@ Returned `status` value is:
 - `INVALID_MEMORY_ACCESS` when `return_serialized_pairs_size` points to
   invalid memory address.
 
+This function is gated on [`HAS_HTTP_HEADERS`], [`HAS_HTTP_WITH_BODY`],
```
leonm1: This is not clear to me if this method is exposed when _any_ of the listed features are enabled or _all_ of the listed features must be enabled to be exposed.

Can you clarify this in the spec for each function that has multiple features listed, to make it explicit?

leonm1 (approved):

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: They could (to avoid crossing Wasm<>Host boundary), but for any unsupported hostcalls, they'll receive `UNIMPLEMENTED` status code anyway, so that's only an optimization.

The main goal is for plugins to verify support for the features they required during configuration phase, and fail to start if some are missing, instead of starting successfully and failing at runtime due to `UNIMPLEMENTED` errors while processing the real requests.

abi-versions/vNEXT/README.md
PiotrSikora: This sentence uses "and/or", which indicates that one, more, or all statements may be true (i.e. it's an "inclusive or"), so it's already explicit, without making it super verbose.

But I use "and/or" quite often, and perhaps it's not as popular as I assumed... How else would you prefer it to be worded?

abi-versions/vNEXT/HOST_FEATURES.md
```diff
@@ -0,0 +1,23 @@
...
+|   0x0003   | `HAS_HTTP_HEADERS`       | When `1`, only baseline functions related to HTTP headers are supported.                           |     No     |
+|   0x0004   | `HAS_HTTP_WITH_BODY`     | When `1`, all baseline functions for HTTP are supported.                                           |     No     |
```
PiotrSikora: @mpwarres @leonm1 do you still need HTTP headers only support?

abi-versions/vNEXT/README.md
leonm1: Thanks for clarifying. I personally am not familiar with this usage of "and/or", but that doesn't mean it can't be used in this fashion.

For spec clarity, I do suggest "This function is implemented if any of the following features enabled: [`HAS_HTTP_HEADERS`], [`HAS_HTTP_WITH_BODY`]", where the if is shorthand for the logical "if and only if".

abi-versions/vNEXT/HOST_FEATURES.md
leonm1: Yes, we will still need HTTP headers without body.

mpwarres: +1, we have some integration points where request/response body is not available.

If we wanted to group HTTP headers + body support in one feature, it wouldn't be the end of the world, but plugins would then need to know that body modification won't always work. So may as well express in a feature, so they have this early indication.

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: Added some text to clarify the ranges.

abi-versions/vNEXT/README.md
PiotrSikora: Done.

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: I'm going back and forth on this. Should we have a generic catch-all text about `UNIMPLEMENTED` status  that applies to all optional hostcalls, or should we explicitly add it to each hostcall... or both?

PiotrSikora: Added text about `NOT_SUPPORTED`.

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: I've split this into `HAS_CALLOUTS_{HTTP,GRPC}_{BUFFERED,STREAMING}`.

PiotrSikora:
@mpwarres @leonm1 I believe all comments and discussion from the meetings have been addressed here (I left the comments open, so please resolve them if you're satisfied with the answers).

leonm1 (approved):
Looks great, thanks!

abi-versions/vNEXT/HOST_FEATURES.md
mpwarres: > Take Envoy for example, it currently supports only fully buffered (i.e. unary) HTTP callouts, so if for some reason we decided to rebuild gRPC support directly on top of HTTP callouts in SDKs, instead of relying on hosts to support gRPC, then we'd lose gRPC streams for callouts.

I think we later concluded that Envoy does support both buffered and streaming HTTP and gRPC callouts.

But no objection to using separate features, since it's possible other proxies might not support both.

abi-versions/vNEXT/HOST_FEATURES.md
mpwarres: Thanks!

abi-versions/vNEXT/HOST_FEATURES.md
mpwarres: The generic catch-all text makes sense to me.

mpwarres (approved):
LGTM, thanks!

leonm1 (approved):
LGTM for merging.

I propose we merge this as-is and handle guest required features in a different PR, so we can build on HOST_FEATURES.md in other PRs.

abi-versions/vNEXT/README.md
```diff
@@ -2105,6 +2325,7 @@ changes to unrelated connections/requests.
 - `UNIMPLEMENTED` = `12`
 - `UNKNOWN_RESOURCE_ID` = `13`
 - `CREATED` = `14`
+- `NOT_SUPPORTED` = `15`
```
PiotrSikora, line 2328: @leonm1 @mpwarres we now have:
- `NOT_SUPPORTED` for unsupported features - either because of missing features, policy, or possibly other reasons,
- `UNIMPLEMENTED` for `proxy_continue_stream(Upstream)` in v0.2.

Do we want to switch that `UNIMPLEMENTED` to more generic `NOT_SUPPORTED`?

PR #92, opened 2025-09-25
Add explicit `context_id` parameter to all hostcalls.

### LINKED ISSUE #59: Context/resource ID should be always provided in the hostcall
Currently, the effective `context_id` is tracked by the host, but this results in a weird asymmetry between hostcalls (without `context_id`) and callbacks (with `context_id`), and makes it harder to build higher-level interfaces on top of those hostcalls.

In order to help with migration from v0.2 to vNEXT, we should accept `context_id=0` as "`context_id` tracked by the host". This should be advertised via settings/feature flag and eventually removed.

PiotrSikora:
While there, add and consistently use `UNKNOWN_RESOURCE_ID` error
for when resources are accessed using an unknown ID.

### DIFF
--- a/abi-versions/vNEXT/HOST_FEATURES.md
+++ b/abi-versions/vNEXT/HOST_FEATURES.md
@@ -1,8 +1,0 @@
-# Host features
-
-| Identifier | Name                            | Values                                                                                             | Deprecated |
-|:----------:|:--------------------------------|:---------------------------------------------------------------------------------------------------|:----------:|
-|   0x1001   | `HOST_TRACKS_ACTIVE_CONTEXT_ID` | When `1`, `context_id` can be set to `0`, which then indicates host-tracked active context ID.     |     Yes    |
-
-
-Identifiers below 0x2000 are reserved for standardized features and options. Random numbers above that range should be used for private extensions.
--- a/abi-versions/vNEXT/HOST_FEATURES.md
+++ b/abi-versions/vNEXT/HOST_FEATURES.md
@@ -1,0 +1,8 @@
+# Host features
+
+| Identifier | Name                            | Values                                                                                             | Deprecated |
+|:----------:|:--------------------------------|:---------------------------------------------------------------------------------------------------|:----------:|
+|   0x1001   | `HOST_TRACKS_ACTIVE_CONTEXT_ID` | When `1`, `context_id` can be set to `0`, which then indicates host-tracked active context ID.     |     Yes    |
+
+
+Identifiers below 0x2000 are reserved for standardized features and options. Random numbers above that range should be used for private extensions.
--- a/abi-versions/vNEXT/README.md
+++ b/abi-versions/vNEXT/README.md
@@ -228,27 +228,6 @@ Returned `status` value is:
 - `NOT_FOUND` when active context was not pending finalization.
 
 
-#### `proxy_set_effective_context`
-
-* params:
-  - `i32 (uint32_t) context_id`
-* returns:
-  - `i32 (`[`proxy_status_t`]`) status`
-
-> **Warning**
-> This function has been deprecated in favor of explicit context IDs.
-
-Changes the effective context to `context_id`.
-
-This can be used to change active context, e.g. during
-[`proxy_on_http_call_response`], [`proxy_on_grpc_receive`]
-and/or [`proxy_on_queue_ready`] callbacks.
-
-Returned `status` value is:
-- `OK` on success.
-- `UNKNOWN_RESOURCE_ID` for unknown `context_id`.
-
-
 ## Configuration
 
 ### Callbacks exposed by the Wasm module
--- a/abi-versions/vNEXT/README.md
+++ b/abi-versions/vNEXT/README.md
@@ -1200,6 +1200,7 @@ Returned `status` value is:
 #### `proxy_http_call`
 
 * params:
+  - `i32 (uint32_t) parent_context_id`
   - `i32 (const char *) upstream_name_data`
   - `i32 (size_t) upstream_name_size`
   - `i32 (const uint8_t *) serialized_headers_data`
@@ -1217,6 +1218,9 @@ Sends HTTP request with [serialized] headers (`serialized_headers_data`,
 `serialized_headers_size`), `body`, and [serialized] trailers
 (`serialized_trailers_data`, `serialized_trailers_size`)
 to upstream (`upstream_name_data`, `upstream_name_size`).
+
+The response will be associated with `parent_context_id` (either
+`plugin_context_id` or `stream_context_id`) and bound to its lifetime.
 
 [`proxy_on_http_call_response`] will be called with `return_call_id`
 when the response is received by the host, or after the `timeout`.
@@ -1272,6 +1276,7 @@ with `map_id` set to `HTTP_CALL_RESPONSE_TRAILERS`.
 #### `proxy_grpc_call`
 
 * params:
+  - `i32 (uint32_t) parent_context_id`
   - `i32 (const char *) upstream_name_data`
   - `i32 (size_t) upstream_name_size`
   - `i32 (const char *) service_name_data`
@@ -1294,6 +1299,9 @@ to gRPC method (`method_name_data`, `method_name_size`)
 on gRPC service (`service_name_data`, `service_name_size`)
 on upstream (`upstream_name_data`, `upstream_name_size`).
 
+The response will be associated with `parent_context_id` (either
+`plugin_context_id` or `stream_context_id`) and bound to its lifetime.
+
 [`proxy_on_grpc_receive`] or [`proxy_on_grpc_close`] will be called
 with `return_call_id` when the response is received by the host, or
 after the `timeout`.
@@ -1315,6 +1323,7 @@ Returned `status` value is:
 #### `proxy_grpc_stream`
 
 * params:
+  - `i32 (uint32_t) parent_context_id`
   - `i32 (const char *) upstream_name_data`
   - `i32 (size_t) upstream_name_size`
   - `i32 (const char *) service_name_data`
@@ -1335,6 +1344,9 @@ on upstream (`upstream_name_data`, `upstream_name_size`).
 
 gRPC messages can be sent on this stream using [`proxy_grpc_send`]
 with `return_stream_id`.
+
+The response will be associated with `parent_context_id` (either
+`plugin_context_id` or `stream_context_id`) and bound to its lifetime.
 
 [`proxy_on_grpc_receive`] or [`proxy_on_grpc_close`] will be called
 with `return_stream_id` when the response is received by the host,
--- a/abi-versions/vNEXT/README.md
+++ b/abi-versions/vNEXT/README.md
@@ -1563,7 +1563,7 @@ Returned `status` value is:
 
 ## Shared Queues
 
-> **Note** 
+> **Note**
 > The default pre-opened shared queue can be accessed using `queue_id=0`.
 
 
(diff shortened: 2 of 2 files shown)


abi-versions/vNEXT/README.md
```diff
@@ -234,6 +235,9 @@ Returned `status` value is:
 * returns:
   - `i32 (`[`proxy_status_t`]`) status`
 
+> **Warning**
+> This function has been deprecated in favor of explicit context IDs.
+
 Changes the effective context to `context_id`.
 
 This can be used to change active context, e.g. during
```
PiotrSikora: Perhaps it should be completely removed?

abi-versions/vNEXT/HOST_FEATURES.md
```diff
@@ -0,0 +1,8 @@
+# Host features
+
+| Identifier | Name                            | Values                                                                                             | Deprecated |
+|:----------:|:--------------------------------|:---------------------------------------------------------------------------------------------------|:----------:|
+|   0x1001   | `HOST_TRACKS_ACTIVE_CONTEXT_ID` | When `1`, `context_id` can be set to `0`, which then indicates host-tracked active context ID.     |     Yes    |
```
PiotrSikora: I originally envisioned this as a helpful step to ease the migration (i.e. SDKs could simply pass `0` as `context_id` and rely on hosts doing the tracking), but I'm not sure if that's worth it?

mpwarres: Would the alternative be for SDKs to track and pass explicit context IDs from the get go, as part of initial support for the new ABI version? Unless you think there are tricky cases where it would be hard for an SDK to track, ISTM it's not too much work for SDKs to do as part of the switchover.

abi-versions/vNEXT/README.md
```diff
@@ -213,17 +213,18 @@ or after a call to [`proxy_done`].
 #### `proxy_done`
 
 * params:
-  - none
+  - `i32 (uint32_t) context_id`
```
mpwarres, line 196: In hostcalls like this one, which list their param as `context_id` rather than `stream_context_id` or `plugin_context_id`, is it correct to interpret that as meaning that either type of context ID can be passed?

For this hostcall in particular, would passing `plugin_context_id` to `proxy_done` be a way of signalling shutdown of the plugin as a whole?

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: Yes, that's the alternative.

I think I was originally worried about plugins directly calling hostcalls, and wanted to avoid adding extra code paths to each hostcall to retrieve SDK-tracked context ID, and instead rely on the existing host-side tracking... But I don't have strong opinion here, and perhaps we don't care about plugins that call the hostcalls directly, so we can drop this.

abi-versions/vNEXT/README.md
PiotrSikora: It == `proxy_set_effective_context`, since we're explicitly passing the context IDs.

Thoughts?

abi-versions/vNEXT/README.md
PiotrSikora: Sorry, I think I was just being lazy here. `proxy_done` can only be called with `plugin_context_id` to signal shutdown of the plugin, that's the only function of this hostcall. It's not supported with `stream_context_id`.

abi-versions/vNEXT/README.md
```diff
@@ -408,22 +412,28 @@ Returned `errno` value is:
 
 ## Timers
 
+> **Note**
+> The default timer can be accessed using `timer_id=0`.
```
PiotrSikora, line 375: This and the default shared KV store (`kvstore_id=0`) and queue (`queue_id=0`) are added to ease migration from single instance model in the current ABI, and to simplify plugins that don't require multiple instances, but perhaps we don't need those default instances?

abi-versions/vNEXT/README.md
leonm1: I see no reason to keep this method around.

It'll still be exported by cpp-host for plugins using older versions of the ABI, it just won't be used by the new ABI version.

abi-versions/vNEXT/HOST_FEATURES.md
leonm1: IMO plugins that break out of the SDK are on their own to get the context ID from the SDK (though we should provide getters for the context id in each context).

abi-versions/vNEXT/README.md
leonm1, line 142: Can you confirm that `context_id` provided here by the host is what is to be passed as a parameter to all the hostcalls?

leonm1 (commented):
My understanding now is that the wasm code itself tells the host which context to use via the context_id parameter. Previously, it was implicit based on the last wasm callback the host invoked.

The context id is given to the wasm code during proxy_on_context_create, then is it the responsibility of the SDK to keep track of it, or is it on plugin authors? It would be helpful guidance to explain this in the description. Can you expand the PR description a bit to express what is changing with this spec change?

abi-versions/vNEXT/README.md
PiotrSikora: Not to all hostcalls, no. Think of it more as a resource ID than a context ID.

This `context_id` would be passed as a parameter to all hostcalls related to the "main" HTTP stream (get headers, set body, etc), but e.g. HTTP/gRPC callouts hostcalls use `callout_id` that's returned from `proxy_dispatch_grpc_call`, K/V store hostcalls use `kvstore_id` that's returned from `proxy_open_shared_kvstore`, etc.

PiotrSikora:
> My understanding now is that the wasm code itself tells the host which context to use via the context_id parameter. Previously, it was implicit based on the last wasm callback the host invoked.

Correct (ignoring the `proxy_set_effective_context` hostcall).

> The context id is given to the wasm code during proxy_on_context_create, then is it the responsibility of the SDK to keep track of it, or is it on plugin authors? It would be helpful guidance to explain this in the description. Can you expand the PR description a bit to express what is changing with this spec change?

Technically, they don't have to be tracked at all, since all callbacks already include resource ID (`context_id`, `callout_id`, `kvstore_id`, etc.)

Having said that, I assume that SDKs would provide objects that track the resource IDs and won't require users to operate directly on them.

As for the guidance, is a bit tricky, because ABI defines Host<>Wasm interface, and whatever happens on the Wasm side is not part of that contract. The plugin authors may or may not use any SDK at all. We could provide some guidance for the host implementations, but I think anything for the SDK users should be documented in the SDKs, since it may vary from SDK to SDK.

abi-versions/vNEXT/HOST_FEATURES.md
mpwarres: It sounds like the consensus is to remove this then?

abi-versions/vNEXT/README.md
mpwarres: In that case, can the param name here be changed to `plugin_context_id` to make that explicit?

abi-versions/vNEXT/README.md
mpwarres: +1 to removing

abi-versions/vNEXT/README.md
PiotrSikora: I'm going to leave it as-is for the time being, since `plugin_on_done`, `plugin_on_log`, `plugin_on_delete` are all passing `context_id` and work with both Plugin and Stream context, so adding this restriction here would be a bit inconsistent (even though it's only used with Plugins in current version of the ABI).

abi-versions/vNEXT/README.md
PiotrSikora: Leaving "always open" resources with `id=0` per [decision in the ABI sync meeting on 2/19](https://docs.google.com/document/d/1M7VTrVaKv2TPuk2Mg9rc8qJGLktgDrU6CERdsumqYEQ/edit?tab=t.0#bookmark=id.ugckx6h2b00y).

abi-versions/vNEXT/README.md
PiotrSikora: Removed.

abi-versions/vNEXT/HOST_FEATURES.md
PiotrSikora: Removed.

abi-versions/vNEXT/README.md
```diff
@@ -2021,6 +2062,7 @@ changes to unrelated connections/requests.
 - `CAS_MISMATCH` = `8`
 - `INTERNAL_FAILURE` = `10`
 - `UNIMPLEMENTED` = `12`
+- `UNKNOWN_RESOURCE_ID` = `13`
```
PiotrSikora, line 2057: MERGE BLOCKER: conflicts with `NOT_SUPPORTED` from #91.

PiotrSikora:
@mpwarres @leonm1 I believe all comments and discussion from the meetings have been addressed here (I left the comments open, so please resolve them if you're satisfied with the answers).

abi-versions/vNEXT/README.md
mpwarres: SGTM

abi-versions/vNEXT/README.md
```diff
@@ -408,22 +391,28 @@ Returned `errno` value is:
 
 ## Timers
 
+> **Note**
+> The default timer can be accessed using `timer_id=0`.
+
+
 ### Functions exposed by the host
 
 #### `proxy_set_tick_period_milliseconds`
```
leonm1, line 381: Does this need to have a context_id?

How does the plugin_context_id get passed to `plugin_context_id`?

abi-versions/vNEXT/README.md
leonm1, line 1181: We talked about cleaning up sidestreams when their parent context is destructed.

Do we want to add a context_id parameter to this hostcall to facilitate this?

leonm1, line 1260: (Same question here)

leonm1, line 1307: (and here)

leonm1 (approved):
Approved, with a few questions on hostcalls without context_id parameters.

abi-versions/vNEXT/README.md
mpwarres, line 39: Consider adding a statement here that the value `0` is commonly used as a null context identifier value. You might even give it a name (e.g. the "null context ID") so it is easy to refer to elsewhere.

mpwarres (approved):

abi-versions/vNEXT/README.md
PiotrSikora: > Does this need to have a context_id?

No, `timer_id` is the top-level object here.

> How does the plugin_context_id get passed to `plugin_context_id`?

Do you mean to `proxy_on_tick`? It doesn't, see: https://github.com/proxy-wasm/spec/pull/93 (I guess this defines the order we need to merge those PRs in...).

PiotrSikora: > No, `timer_id` is the top-level object here.

Thinking about this, perhaps we should refer to those as "resource ID"?

abi-versions/vNEXT/README.md
PiotrSikora: Done (as `parent_context_id`).

PiotrSikora: Done (as `parent_context_id`).

PiotrSikora: Done (as `parent_context_id`).

leonm1 (approved):
LGTM for merging.

abi-versions/vNEXT/README.md
PiotrSikora: We're using `0` in a few places where it has different meaning - `parent_context_id=0` refers to "VM context" and `queue_id=0`, `kvstore_id=0`, and `timer_id=0` refer to "the default / global resource", so I'd rather not have common text to describe those.

PR #94, opened 2025-10-09
Document how proxy_{get,set}_property expects paths to be serialized.

mpwarres:
Document the de-facto serialization format used by the various proxy-wasm SDKs ([C++](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/e5256b0c5463ea9961965ad5de3e379e00486640/proxy_wasm_api.h#L931-L943), [Rust](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/ba23ab402e3c2a1fcaa9592d8204333d7b6ae46f/src/hostcalls.rs#L1211-L1226), [Go](https://github.com/proxy-wasm/proxy-wasm-go-sdk/blob/ab4161dcf9246a828008b539a82a1556cf0f2e24/proxywasm/internal/serde.go#L78-L96)) to serialize `path_data` arguments for the `proxy_get_property` and `proxy_set_property` hostcalls.

### DIFF
--- a/abi-versions/v0.2.1/README.md
+++ b/abi-versions/v0.2.1/README.md
@@ -1710,6 +1710,8 @@ Returned `status` value is:
 Retrieves value (`return_value_data`, `return_value_size`)
 of the property (`path_data`, `path_size`).
 
+`path_data` is a [serialized] list of path segments.
+
 Returned `status` value is:
 - `OK` on success.
 - `NOT_FOUND` when there was no property found at the requested `path`.
@@ -1731,6 +1733,8 @@ Returned `status` value is:
 
 Sets value of the property (`path_data`, `path_size`) to the provided
 value (`value_data`, `value_size`).
+
+`path_data` is a [serialized] list of path segments.
 
 Returned `status` value is:
 - `OK` on success.
@@ -1943,6 +1947,14 @@ e.g. the map `{{"a": "1"}, {"b": "22"}}` would be serialized as:
 An empty map may be represented either as an empty value (`size=0`), or as
 a single `0x00` byte (`size=1`).
 
+#### Property path names
+
+Path data for the [proxy_get_property] and [property_set_property] hostcalls
+consists of a sequence of path segments. The path segments are null-terminated
+and then concatenated in sequence to form the `path_data` argument.
+
+e.g. the path segments `["foo", "bar"]` would be serialized as:
+-  `0x66`, `0x6f`, `0x6f`, `0x00`, `0x62`, `0x61`, `0x72`, `0x00`
 
 # Security Considerations
 
--- a/abi-versions/v0.2.1/README.md
+++ b/abi-versions/v0.2.1/README.md
@@ -1959,6 +1959,7 @@ e.g. the path segments `["foo", "bar"]` would be serialized as:
 Host implementations should tolerate a `NULL` character at the end of the
 combined path data string, if present.
 
+
 # Security Considerations
 
 ## External resources
--- a/abi-versions/v0.2.1/README.md
+++ b/abi-versions/v0.2.1/README.md
@@ -1950,11 +1950,14 @@ a single `0x00` byte (`size=1`).
 #### Property path names
 
 Path data for the [proxy_get_property] and [property_set_property] hostcalls
-consists of a sequence of path segments. The path segments are null-terminated
-and then concatenated in sequence to form the `path_data` argument.
+consists of a sequence of path segments. The path segments are separated by
+`NULL` (`0x00`) characters.
 
 e.g. the path segments `["foo", "bar"]` would be serialized as:
--  `0x66`, `0x6f`, `0x6f`, `0x00`, `0x62`, `0x61`, `0x72`, `0x00`
+-  `0x66`, `0x6f`, `0x6f`, `0x00`, `0x62`, `0x61`, `0x72`
+
+Host implementations should tolerate a `NULL` character at the end of the
+combined path data string, if present.
 
 # Security Considerations
 
--- a/abi-versions/v0.2.0/README.md
+++ b/abi-versions/v0.2.0/README.md
@@ -1687,6 +1687,8 @@ Returned `status` value is:
 Retrieves value (`return_value_data`, `return_value_size`)
 of the property (`path_data`, `path_size`).
 
+`path_data` is a [serialized] list of path segments.
+
 Returned `status` value is:
 - `OK` on success.
 - `NOT_FOUND` when there was no property found at the requested `path`.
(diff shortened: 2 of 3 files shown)


abi-versions/v0.2.1/README.md
```diff
@@ -1943,6 +1947,14 @@ e.g. the map `{{"a": "1"}, {"b": "22"}}` would be serialized as:
...
 a single `0x00` byte (`size=1`).
 
+#### Property path names
+
+Path data for the [proxy_get_property] and [property_set_property] hostcalls
+consists of a sequence of path segments. The path segments are null-terminated
+and then concatenated in sequence to form the `path_data` argument.
+
+e.g. the path segments `["foo", "bar"]` would be serialized as:
+-  `0x66`, `0x6f`, `0x6f`, `0x00`, `0x62`, `0x61`, `0x72`, `0x00`
```
PiotrSikora: That's not entirely correct. There is no `NULL` character at the end.

abi-versions/v0.2.1/README.md
```diff
@@ -1943,6 +1947,14 @@ e.g. the map `{{"a": "1"}, {"b": "22"}}` would be serialized as:
 An empty map may be represented either as an empty value (`size=0`), or as
 a single `0x00` byte (`size=1`).
 
+#### Property path names
+
+Path data for the [proxy_get_property] and [property_set_property] hostcalls
+consists of a sequence of path segments. The path segments are null-terminated
```
PiotrSikora: Segments are separated by a `NULL` (`0x00`) character, not terminated by it.

PiotrSikora (changes requested):
cc @kyessenov

abi-versions/v0.2.1/README.md
mpwarres: Hmm, while you're right about the Rust and Go SDKs not including a `NULL` char at the end, the C++ SDK does: [code](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/e5256b0c5463ea9961965ad5de3e379e00486640/proxy_wasm_api.h#L905-L916).

I can fix the C++ SDK to trim the last `NULL` char, but I think that host implementations would still need to tolerate a terminating `NULL` char in order to work with C++ plugins compiled prior to this fix. The Envoy implementation of `getProperty` does appear to [tolerate](https://github.com/envoyproxy/envoy/blob/6ab1e7afbfda48911e187c9d653a46b8bca98166/source/extensions/common/wasm/context.cc#L551-L560) this.

ISTM that in order to document the acceptable range of behaviors, the spec should say that `path_data` can (but is not required to) contain a terminating `NULL`char. WDYT?

mpwarres: Updated to state that null char is a separator, but host implementations should also tolerate it as a terminator if present.

abi-versions/v0.2.1/README.md
mpwarres: Ack, see response to other comment, and update.

abi-versions/v0.2.1/README.md
kyessenov: The segments could be empty, so the distinction is important actually. Although I don't recall if we ever tested it with empty segments, so it's under-specified.

mpwarres: My reading of [this logic](https://github.com/envoyproxy/envoy/blob/ce94a616479be89f47b440a688fa6c8cbceb570d/source/extensions/common/wasm/context.cc#L549-L560) in `Envoy::Extensions::Common::Wasm::getProperty()` is that:

1. It handles each segment by finding the next null char (if any), and taking the substring up until that (which could be empty)
2. It bumps `start` to the previous `end` (which doesn't include null char) + 1 (skipping over the null char, if present)
3. Loop terminates when `start >= path.size()`. In the case of a `path` that does include a terminating null char, `start` will end up `== path.size()`, while in the case of a `path` that doesn't, `start` will end up equal to `path.size() + 1`. Either way, the loop terminates.

So I think it's still accurate to say that null char is a separator (which doesn't rule out empty segments), and that an optional terminating null char should be tolerated by host implementations.

If there's a different wording that either of you would prefer, I'm open to suggestions. Thanks!

PiotrSikora (commented):
I think we should also add this to `v0.2.0` and `vNEXT` (since any changes from `v0.2.1` should be done via PRs and not by using a stale copy of the spec).

abi-versions/v0.2.1/README.md
PiotrSikora: The current text works fine for me, thanks!

To be honest, I'm actually surprised that the trailing `\0` works fine, since I recall having issues with it, hence the extra code to remove it in Rust & Go SDKs... But I was also under the impression that the serialization format came from CEL and wasn't our own invention, so my memory is clearly a little hazy around those details.

kyessenov: The binary format didn't come from CEL. The constraint was that "." wasn't good enough since segments can contain ".". But you should feel free to revise for vNext with a more standard list packing format.

mpwarres:
> I think we should also add this to `v0.2.0` and `vNEXT` (since any changes from `v0.2.1` should be done via PRs and not by using a stale copy of the spec).

Right--done.

abi-versions/v0.2.1/README.md
mpwarres: > The current text works fine for me, thanks!

Great!

> To be honest, I'm actually surprised that the trailing `\0` works fine, since I recall having issues with it, hence the extra code to remove it in Rust & Go SDKs... But I was also under the impression that the serialization format came from CEL and wasn't our own invention, so my memory is clearly a little hazy around those details.

Yeah, I'm not sure--maybe something changed over time? FWIW my reading of the [current logic](https://github.com/envoyproxy/envoy/blob/6ab1e7afbfda48911e187c9d653a46b8bca98166/source/extensions/common/wasm/context.cc#L551-L560) in `Context::getProperty()` is that (currently) the trailing '\0' would never even be seen by any of the CEL-evaluation functions.

I'll draft a PR for the C++ SDK to stop including it, for consistency with the other SDKs.

mpwarres: > The binary format didn't come from CEL. The constraint was that "." wasn't good enough since segments can contain ".". But you should feel free to revise for vNext with a more standard list packing format.

Ah, TIL! I had wondered about that, interesting to know the history.

mpwarres:
@PiotrSikora PTAL again--thanks!

PiotrSikora (approved):
LGTM, thanks!

Please capitalize the commit message summary & add trailing dot before merging.

mpwarres:
Thanks! Done.

PR #97, opened 2026-02-12
Combine proxy_{register,resolve}_shared_queue into proxy_open_shared_queue.

PiotrSikora:
While there, add proxy_delete_shared_queue.

### DIFF
--- a/abi-versions/vNEXT/README.md
+++ b/abi-versions/vNEXT/README.md
@@ -1556,7 +1556,8 @@ with `return_queue_id`.
 Returned `status` value is:
 - `OK` on success when opening existing queue.
 - `CREATED` on success when a queue with the given name was created.
-- `NOT_FOUND` when the requested queue was not found.
+- `NOT_FOUND` when `create_if_not_exist` is `false` and no queue with
+  the given name exists.
 - `INVALID_MEMORY_ACCESS` when `vm_id_data`, `vm_id_size`, `name_data`,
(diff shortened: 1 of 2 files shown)


PiotrSikora:
Stacked on #92, review only 3edd9a9842555a37ba19522e5d65403a7dc666ad.

leonm1 (approved):

abi-versions/vNEXT/README.md
```diff
@@ -1569,43 +1569,20 @@ Returned `status` value is:
 
 ### Functions exposed by the host
 
-#### `proxy_register_shared_queue`
+#### `proxy_open_shared_queue`
```
leonm1, line 1538: I think we decided that we wanted to add a return argument to indicate if the queue already existed or was created (to permit one-time initialization scripts to run).

From the March 19th proxy-wasm meeting.

mpwarres: Same comment as on #98, an alternative to an added bool return argument would be adding an ALREADY_EXISTS enum value to proxy_status_t.

mpwarres (approved):

abi-versions/vNEXT/README.md
PiotrSikora: Added `CREATED`, since `ALREADY_EXISTS` sounds like an error, and this is a successful code path.

Note that we don't have an error path for creating a queue with the name that's already in use, neither in v0.2 nor with this PR.

PR #98, opened 2026-02-12
Add support for multiple key-value stores.

### LINKED ISSUE #58: Add new/delete for most/all resources

PiotrSikora:


PiotrSikora:
Stacked on #92, review only 755dff1c1102de3233b13e9e19f0f687e2c2508b.

leonm1 (approved):

abi-versions/vNEXT/README.md
```diff
@@ -1506,6 +1506,33 @@ gRPC status message can be retrieved using [`proxy_get_status`].
 
 ### Functions exposed by the host
 
+#### `proxy_open_kvstore`
+
+* params:
+  - `i32 (const char*) kvstore_name_data`
+  - `i32 (size_t) kvstore_name_size`
```
leonm1, line 1478: I think we decided that we wanted to add a return argument to indicate if the kvstore already existed or was created (to permit one-time initialization scripts to run)

From the March 19th proxy-wasm meeting.

mpwarres: Maybe we could add an ALREADY_EXISTS value to the proxy_status_t enum?

leonm1: I think a `CREATED` status, to mirror HTTP 201 would be good

PiotrSikora: Added `CREATED`, since `ALREADY_EXISTS` sounds like an error, and this is a successful code path.

Note that we don't have an error path for creating a key-value store with the name that's already in use.

PR #99, opened 2026-02-12
Document `proxy_get_shared_data(key, NULL)` (exists).

### LINKED ISSUE #70: KVS: add ability to check existence of a given key without fetching it
We need a way to check existence of a given key without extra copies.

### LINKED ISSUE #80: Consistent style in the ABI (expanded vs minimal)
During the initial development, we went through a lot of back-and-forth both expanding and minimizing the ABI surface (e.g. `get_buffer_bytes(buffer_type)` vs `get_http_request_body_chunk()` and similar, `set_shared_data(NULL)` vs `remove_shared_data()`), both internally and during Envoy upstreaming process.

The version that was initially merged into Envoy, and which the ecosystem ossified around, ended up with some unfortunate inconsistencies (using both: expanded `proxy_{get,add,replace,remove}_header_map_value`, and minimal `proxy_set_buffer_bytes(start, offset)` or `proxy_set_shared_data(NULL)`).

We should clean this up and stick to one style.

While I don't have strong preference one way or the other, note that splitting features into groups (#57) and having specialized context types with their own hostcalls (#9) probably means that we should be expanding the ABI surface instead of minimizing it.

@mpwarres @martijneken @leonm1 @wbpcode @thibaultcha any thoughts?

PiotrSikora:


leonm1 (approved):

leonm1 (approved):
LGTM for merging

PR #101, opened 2026-03-19
Remove `proxy_remove_header_map_value`.

### LINKED ISSUE #80: Consistent style in the ABI (expanded vs minimal)

PiotrSikora:


leonm1 (approved):

abi-versions/vNEXT/README.md
leonm1: I didn't see a PR to rename this to proxy_set_header_map_value (decided in March 19th meeting). Do you want to do that here, or in a separate PR?

mpwarres: Might be simpler to just do it here, but either way SGTM.

mpwarres (approved):

abi-versions/vNEXT/README.md
PiotrSikora: Included here.

PR #104, opened 2026-05-21
Remove `num_headers` and use `status_code` in responses instead.

### LINKED ISSUE #28: Is the num_headers argument in used?
Since now we fetch HTTP headers as a map, do we need a num_headers argument？

PiotrSikora:


abi-versions/vNEXT/README.md
```diff
@@ -1214,30 +1213,35 @@ Returned `status` value is:
 * params:
   - `i32 (uint32_t) plugin_context_id`
   - `i32 (uint32_t) call_id`
-  - `i32 (size_t) num_headers`
+  - `i32 (size_t) status_code`
   - `i32 (size_t) body_size`
-  - `i32 (size_t) num_trailers`
+  - `i32 (size_t) has_trailers`
```
leonm1, line 1243: Does proxy_get_header_map_pairs support "has" semantics by passing NULL for a return buffer pointer? If so, we don't need `has_trailers`.

PiotrSikora: The point of those hints is to avoid the expensive call across the host<>Wasm boundary when not needed.

leonm1 (approved):

PR #105, opened 2026-06-09
Add upstream selection support

Lakyn03:
Added proposal callbacks and hostcalls enabling plugins to:
- choose the upstream peer for HTTP requests
- configure per-request timeouts
- handle whether response with error status codes (4xx, 5xx) should be forwarded downstream or another attempt should be made
- store the outcome of the last attempt to contact an upstream server

Prototype implementation for Nginx:
https://github.com/Kong/ngx_wasm_module/pull/710


PiotrSikora:
@Lakyn03 we cannot look at the content of this PR until the CLA is signed.

